PR #17, opened 2023-01-05
update help files of pdlua/pdluax

porres:
please revise and see if you like it. This works with the way things are currently shipped in the current release downloads.


agraef:
These are indeed much nicer, thanks!

PR #19, opened 2023-01-05
fix mistakes in the pdlua help file and change color theme

porres:
There were errors in pdlua because I moved things around, so I was asking people to look below or above when it was the other way around.

I also couldn't hold myself and went ahead and changed the color scheme to something less 'colorful' cause those shiny colors always burn my eyes and I can hardly read them (I hope you don't mind). This is for the 'howto' sections. I went for a more graysh look I am used to but can try other things. I also took advantage of the fact that you now can use spaces in canvas' labels.


agraef:
Hmm, I didn't notice much change in the colors of the HOWTO subpatches, but maybe they're too subtle for me to notice.

PR #25, opened 2023-01-09
Fixed menu_openfile compatibility with plugdata

timothyschoen:


agraef:
Thanks! Progress -- the Open option isn't greyed out any more. :))

The directory is the wrong one, though. class->c_externdir->s_name doesn't refer to the pdlua external and dir, it's the dir of the *Lua object class* at hand for which we search the .pd_lua source file.

But at least I better understand the issue now. I must check where the c_externdir actually gets set when the object gets registered on the Lua side, and why that path always comes up empty. I can certainly work around this issue on the pdlua side because the full path *is* known at some point when the loader loads the script into the Lua interpreter. But there's probably some API missing or not implemented in the same way in plugdata/libpd as well, otherwise this code should work as it does in vanilla and purr.

I will look into it.

agraef:
Ok, I see. The thing is that class->c_externdir *is* now set for proper Lua objects with .pd_lua files. Whatever you did there, this works now. :)) Your PR will work for the pdlua object itself, so I just need to combine the two. Stay tuned, you'll get a PR with another pd-lua update real soon.

Issue #30, opened 2023-04-11
Fix Makefile installplus dependency


a1ca7raz:
I am packaging this project for Arch Linux RISC-V and I noticed that *make install* doesn't work on my riscv64 platform.
```
make: Entering directory '/build/pd-lua/src/pd-lua-0.11.6'
++++ NOTE: using installed lua
++++ info: using Makefile.pdlibbuilder version 0.6.0
++++ info: making target install in lib pdlua
++++ info: installpath is '/build/pd-lua/pkg/pd-lua//usr/lib/pd/extra/pdlua'
++++info: target all in lib pdlua completed
++++ info: executables of lib pdlua installed from /build/pd-lua/src/pd-lua-0.11.6 to /build/pd-lua/pkg/pd-lua//usr/lib/pd/extra/pdlua
install -m 755 -d -v "/build/pd-lua/pkg/pd-lua//usr/lib/pd/extra/pdlua"
++++ info: data files of lib pdlua installed from /build/pd-lua/src/pd-lua-0.11.6 to /build/pd-lua/pkg/pd-lua//usr/lib/pd/extra/pdlua
install -m 755 -d -v "/build/pd-lua/pkg/pd-lua//usr/lib/pd/extra/pdlua"
cp -r ./pdlua/ "/build/pd-lua/pkg/pd-lua//usr/lib/pd/extra/pdlua"/pdlua
/usr/bin/install: creating directory '/build/pd-lua/pkg/pd-lua//usr'
/usr/bin/install: creating directory '/build/pd-lua/pkg/pd-lua//usr/lib/pd'
/usr/bin/install: creating directory '/build/pd-lua/pkg/pd-lua//usr/lib/pd/extra'
/usr/bin/install: creating directory '/build/pd-lua/pkg/pd-lua//usr/lib/pd/extra/pdlua'
/usr/bin/install: creating directory '/build/pd-lua/pkg/pd-lua//usr/lib'
install -p -m 644 'pdlua.pd_linux' "/build/pd-lua/pkg/pd-lua//usr/lib/pd/extra/pdlua";
/usr/bin/cp: cannot create directory '/build/pd-lua/pkg/pd-lua//usr/lib/pd/extra/pdlua/pdlua': No such file or directory
[...]

PR #31, opened 2023-04-11
fix installplus dep

a1ca7raz:
see https://github.com/agraef/pd-lua/issues/30

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,5 +48,8 @@ include Makefile.pdlibbuilder
 
 install: installplus
 
+installplus: $(if $(executables), install-executables)
+installplus: $(if $(datafiles), install-datafiles)
+installplus: $(if $(datadirs), install-datadirs)
 installplus:
 	cp -r ./pdlua/ "${installpath}"/pdlua


agraef:
Sorry for getting back so late, I guess I overlooked your PR during busy times.

#30 seems to arise because I assumed that the prerequisites of install in Makefile.pdlibbuilder would be made before the installplus requisite added in the Makefile. Which obviously isn't right, because the requisites can be made in any order, and even in parallel.

But your solution depends on these internal prerequisites in Makefile.pdlibbuilder, which might break due to upstream changes if Makefile.pdlibbuilder gets updated (which I do from time to time, by pulling the latest version from upstream).

I think that the proper solution would be to just add `$(INSTALL_DIR) -v "$(installpath)"` to installplus, i.e.:

~~~make
installplus:
	$(INSTALL_DIR) -v "$(installpath)"
	cp -r ./pdlua/ "${installpath}"/pdlua
~~~

By these means there's only a single potential point of breakage, `$(INSTALL_DIR)`, and that's unlikely to change.

Can you try whether that works for you and, if so, augment the PR accordingly? Thanks.

agraef:
I've just added my proposed fix to the master branch, can you please test that it works for you? Then we can close this PR.

Issue #32, opened 2023-07-03
please make pdlibbuilder overridable


umlaeute:
as documented in https://github.com/pure-data/pd-lib-builder/blob/master/tips-tricks.md#project-management

this helps me tremendously when packaging for Debian

PR #33, opened 2023-07-03
Make path to Makefile.pdlibbuilder settable

umlaeute:
Closes: https://github.com/agraef/pd-lua/issues/32

the 2nd commit excludes the `.gitmodules` file  (and `.github/` directory) from the generated source tarballs.
(the reasoning is: if you want all the git-nifties of this repository, you should probably just clone the repository; but if you are using the source-tarballs for downstream packaging *and* `git` to manage the packaging, upstream repository config usually gets in the way badly)

the 3rd commit allows packagers to override the `pkg-config` script, which helps when cross-compiling.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+.git*      	export-ignore
+.travis.yml	export-ignore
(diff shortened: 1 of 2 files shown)


agraef:
Merged, thanks!

PR #40, opened 2024-03-02
Windows fixes for randomness and hi-dpi scaling

### LINKED ISSUE #39: graphics api results in tcl errors in console on Windows
On windows 10 (Tcl/tk 8.6.13 Pd 0.54-1) when using the introductory help file and the new gui example there are errors in the console:
```
(Tcl) UNHANDLED ERROR: unknown option "-font"
    while executing
".x71416d90.c itemconfigure {.x13146} -font {{DejaVu Sans Mono} 8 {bold} } -fill {#FC7651} -justify {left} "
    ("uplevel" body line 53)
    invoked from within
"uplevel #0 $docmds"
```
the canvas item tag varies and sometimes only has 3 or 4 numerical digits. (e.g. ```{.x888}```)
this also happens for lines 35, 38, 41, 53, 56, 104

there is also
```
(Tcl) UNHANDLED ERROR: wrong # coordinates: expected 0 or 4, got 22
    while executing
".x71416d90.c coords {.x16505} 318 222 323 236 338 236 326 244 332 258 318 249 304 258 312 244 300 236 315 236 318 222 "
    ("uplevel" body line 26)
    invoked from within
"uplevel #0 $docmds"
```

timothyschoen:
Fixes 2 problems on Windows:

- Better random string generation, which fixes #39. We now have a string with 61 random alphanumeric chars (also using a better random algorithm), which should give plenty of randomness.

- Fontheight needs to be passed to tcl/tk as a negative value to get it to scale along with hi-dpi screens on Windows. This is done in other places in pd's source code as well, like [here](https://github.com/pure-data/pure-data/blob/0259d00373b9e6f3bc1d76eebc5b243283ca7063/src/g_all_guis.c#L543)

### DIFF
--- a/pdlua.h
+++ b/pdlua.h
@@ -31,9 +31,9 @@ typedef struct _pdlua_gfx
 {
 #if !PLUGDATA
     char object_tag[128]; // Tcl/tk tag that is attached to all drawings
-    char order_tag[128]; // Tag for invisible line, used to preserve correct object ordering
-    char current_paint_tag[128]; // Tcl/tk tag that is only attached to the current drawing in progress
-            
+    char order_tag[64]; // Tag for invisible line, used to preserve correct object ordering
+    char current_paint_tag[64]; // Tcl/tk tag that is only attached to the current drawing in progress
+    
     // Variables to keep track of mouse button state and drag position
     int mouse_drag_x, mouse_drag_y, mouse_down;
     
--- a/pdlua_gfx.h
+++ b/pdlua_gfx.h
@@ -1008,10 +1008,12 @@ static int draw_text(lua_State* L) {
     int x = luaL_checknumber(L, 2);
     int y = luaL_checknumber(L, 3);
     int w = luaL_checknumber(L, 4);
-    int fontHeight = luaL_checknumber(L, 5);
+    int font_height = luaL_checknumber(L, 5);
+    font_height = sys_hostfontsize(font_height, glist_getzoom(cnv));
+    font_height *= 0.75f; // Make font size smaller to match the size in plugdata
     
     transform_point(gfx, &x, &y);
-    transform_size(gfx, &w, &fontHeight);
+    transform_size(gfx, &w, &font_height);
     
     int canvas_zoom = glist_getzoom(cnv);
     x += text_xpix((t_object*)obj, obj->canvas) / canvas_zoom;
@@ -1021,9 +1023,6 @@ static int draw_text(lua_State* L) {
     y *= canvas_zoom;
     w *= canvas_zoom;
     
-    // Font size is offset to make sure it matches the size in plugdata
-    fontHeight *= 0.8f;
-    
     const char* tags[] = { gfx->object_tag, register_drawing(obj) };
     
     pdgui_vmess(0, "crr ii rs ri rs rS", cnv, "create", "text",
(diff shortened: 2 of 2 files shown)


timothyschoen:
I still have one small PR coming up. The way I currently store the currently active object in the Lua registry is hacky and needs to be fixed. It could also potentially cause bugs.

I can't get rid of it entirely without changing the API, but I think I can at least minimise its usage as much as possible.

sebshader:
the errors seem to be gone now.
while the font in pd for windows already looks somewhat small, is it supposed to be this small? it's kind of difficult to read here (though maybe my font/display settings are j
pdlua-font
ust too small overall..)

timothyschoen:
> the errors seem to be gone now. while the font in pd for windows already looks somewhat small, is it supposed to be this small? it's kind of difficult to read here (though maybe my font/display settings are just too small overall somehow..) pdlua-font
> 
> edit: I see the font is scaled down to match plug-data, is it that small in plug-data too?

Yeah, I guess the problem is that the text is set too small in the example, since you can set your own font size. It's also too small in plugdata

agraef:
Indeed, the example uses 10pt for the text, which is really small on hi-dpi displays. Also, the yellow foreground color on the pink background make the Bezier example just barely visible, at least for my eyes. We should maybe change that in hello-gui.pd_lua. E.g., here is how it looks like with a dark gray background and the font size set to 12 pt, which also seems to be the default in new patch windows in vanilla:

image

I think that this looks a lot better. Obviously, I still need to change the text a bit to make everything fit again. But first let me merge this PR so that those annyoing bugs on Windows are fixed.

agraef:
@timothyschoen I don't think that the new factor of 0.75 applied at line 1013 is correct. After merging this, the font sizes are even smaller than before, at least on Linux (this is with desktop scaling at 100%, but on Linux the desktop scale doesn't seem to affect the canvas anyway, only the GUI):

image

Everything is back to normal after commenting out line 1013:

image

So I think that we need to get rid of that line. Maybe you could adjust the scaling on the plugdata side instead? Or adjust the scaling depending on the OS and the desktop scaling that is in effect? Gets complicated, I know, but the way it is now, the font sizes are way off in vanilla and 10 is completely illegible.

agraef:
Ok, here's how the demo looks now after fiddling with some of the positions, and after commenting out line 1013:

image

I've included the [osc~] object for reference, with the canvas font size set to 12, which matches the font size in the gfx example. The font sizes look alright to me.

I'm ready to commit this so that you can test it on vanilla in Windows, and on plugdata. Opinions?

agraef:
Here's the example that I used for testing, with the modified hello-gui.pd_lua included. But note that you'll need to comment out line 1013 and recompile to compare with the screenie that I posted above.

[gfx-test.zip](https://github.com/agraef/pd-lua/files/14472224/gfx-test.zip)

agraef:
Ok, I just went ahead and commited my changes now, so that you can test. @timothyschoen I'm ready to revert the font scaling change (rev. d7f6fe8875b1a38f458d08214c446634efcc6440) as soon as you come up with something better.

timothyschoen:
> Ok, I just went ahead and commited my changes now, so that you can test. @timothyschoen I'm ready to revert the font scaling change (rev. [d7f6fe8](https://github.com/agraef/pd-lua/commit/d7f6fe8875b1a38f458d08214c446634efcc6440)) as soon as you come up with something better.

Screenshot 2024-03-03 at 16 56 09

I've now just scaled the fontsize on plugdata's side so it's about the same as in pd-vanilla. I removed your comment in #42 as well.

PR #49, opened 2024-08-25
fix deken upload

ben-wes:
upload job should work now like in the previous version.


agraef:
Thanks! So let's give this another go...

ben-wes:
phew ... looks correct again now! sorry for the trouble here ... i guess i can learn from this. ;)

PR #50, opened 2024-08-25
Fix alpha colour support in plugdata

timothyschoen:
I think this should be possible in purr-data too? I added this, since there is basically no other way to do transparency in Pd.


agraef:
Ah, I already wondered why I couldn't get transparency in plugdata to work. Not sure whether Tk supports alpha blending nowadays, but even if it does, vanilla Pd probably doesn't.

> I think this should be possible in purr-data too?

Sure is, the GUI is basically Chromium after all:

purr-data-alpha

Note that both the comment text and the canvas rectangle are actually *behind* the dial, which a pdlua object. All parts of the pdlua object (except the border of the dial's face) have various degrees of transpancy.

The required changes in pdlua_gfx.h and pdlua.c are minimal. I'm ready to commit them along with your PR.

agraef:
Merged, thanks!

agraef:
BTW, what's up with those palette colors 0, 1, 2? They seem to be different between plugdata and vanilla where both 0 and 2 are white, and 1 is black. Should any of these have transparency in plugdata and purr-data?

PR #51, opened 2024-08-27
minor workflow simplifications

ben-wes:
since pd-lua is using the actual tag name as version string, there's no need for extracting it from another string.

and the check for a tag is also simpler this way.

tested with https://github.com/ben-wes/pd-lua/actions/runs/10574258362

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -97,7 +97,7 @@ jobs:
 # Big thanks to @tomara-x and @timothyschoen for showing me how to do this! -ag
 
   github-release:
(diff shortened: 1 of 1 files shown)


agraef:
Merged, thanks! I'll test this with a proper release asap. What could possibly go wrong? :smile_cat:

ben-wes:
> Merged, thanks! I'll test this with a proper release asap. What could possibly go wrong? 😸

exactly my though, haha - just a little PR for the thrill. ;)

(on a more serious note: not planning to mess much with those workflows in the future unless necessary for actual new features)

Issue #52, opened 2024-09-15
handling of dollar args


ben-wes:
this is more of a discussion since it's not a certain issue and probably there are different ways of handling this. therefore the rather generic title.

---

... the more i look into this, the more complex it seems to become for me. one rather simple use case might be:

* user wants to make his pd_lua object's receive name configurable through a message

this can be easily done with a mechanism like (omitting all checks here):
```lua
function test:in_1_receive(x)
  self.recv:destruct()
  self.recv = pd.Receive:new():register(self, x[1], "receive")
end
```

it becomes more challenging though if there are dollar arguments involved. a typical option would be a message like `receive \$0-foo` to set the name to `$0-foo`, where `$0` gets internally expanded to the actual canvas id. this is the mechanism that's common for Pd's gui objects like toggle, sliders etc.

if i send `receive \$0-bla` to the mechanism above, the dollar argument isn't expanded, which is probably expected, since `$0` is obviously dynamic and shouldn't be forwarded in its expanded form, which would negate its purpose. but in that case, some kind of `canvas_realizedollar()`[^1] function would be needed on the lua side, i assume?

when i use `set_args()` with the received symbol, it ends up in the patch file as `\\\$0-foo` - which means that the backslash gets escaped as well. might be expected, but prevents Pd from properly expanding the symbol on load.

attaching a pdlua object and patch for testing here[^2] - the updated arguments of [test_dollar] are obviously only visible on reload or cut/paste.

[^1]: https://github.com/pure-data/pure-data/blob/master/src/g_canvas.c#L279-L292
[^2]: [test_dollar.zip](https://github.com/user-attachments/files/17005195/test_dollar.zip)

agraef:
Well, canvas_realizedollar is in g_canvas.h which, at least at the time of this writing, is among the public header files, so in principle it would be possible to expose it in pd-lua's Lua API in some way.

But I don't really see why you need to to expand the symbol on the Lua side when Pd can do it just fine. Admittedly, the [bang( - [f $0] - [receive $1-foo( - [your Lua object] construct is a bit clunky, but it's what Miller's mandates in his html manual. (Incidentally, the clumsiness of this construct is probably the reason why in Pd-l2ork / Purr Data we can just write [receive $0-foo( and be done with it.)

Regarding the extra level of escapes in set_args, I'm sure that this is not in set_args itself, so it must happen inside Pd's guts (probably somewhere in the binbuf operations). Alas, I don't think that there's much that we can do about that. :(

[...]

PR #54, opened 2024-09-18
expose Pd's canvas_realizedollar() and modify set_args() for dollar arguments

ben-wes:
creating this PR here mainly to summarize the result of https://github.com/agraef/pd-lua/issues/52, which escalated a bit. these changes certainly require review, and it's obviously also no problem if this PR is declined or needs further modifications.

#### the primary goal was:
* enable pdlua externals to have their senders and receivers set through messages, similar to iemguis and other externals.

#### changes:
1. writing the necessary `\$0-name` string to the pd file isn't possible with `set_args()` as it also escapes the backslash to `\\\$0-name`. this is addressed with 48f00537b04ef2100e10c7e52aaec877b4b32d0d
2. for the pdlua object to apply the given receiver name immediately, an API method to expand it is required. this functionality is implemented with 53f0342b1db016c5e1112e10b63532c4f5e7aefb

#### findings that may require further consideration:
1. when a dollar argument can't be expanded, it remains a dollar argument (in lua, it will just be a `"$1"` string then, for example)
2. sending `\$01` from pd in that case will also result in `"$1"` on the lua side however

i've tested these changes in my setup without encountering issues so far. to facilitate testing elsewhere, i'm attaching these files (pdlua object and patch): [test_dollar.zip](https://github.com/user-attachments/files/17049903/test_dollar.zip)

### DIFF
--- a/pd.lua
+++ b/pd.lua
@@ -453,6 +453,10 @@ function pd.Class:set_args(args)
     pd._set_args(self._object, args)
 end
 
+function pd.Class:canvas_realizedollar(s)
+    return pd._canvas_realizedollar(self._object, s)
+end
+
 function pd.Class:repaint()
   if type(self.paint) == "function" then
     local g = _gfx_internal.start_paint(self._object);
--- a/pdlua.c
+++ b/pdlua.c
@@ -2490,6 +2490,23 @@ static int pdlua_dofile(lua_State *L)
     return lua_gettop(L) - n;
 }
 
+static int pdlua_canvas_realizedollar(lua_State *L)
+{
+    if (lua_islightuserdata(L, 1) && lua_isstring(L, 2))
+    {
+        t_pdlua *o = lua_touserdata(L, 1);
+        if (o && o->canvas)
+        {
+            const char *sym_name = lua_tostring(L, 2);
+            t_symbol *s = gensym(sym_name);
+            t_symbol *result = canvas_realizedollar(o->canvas, s);
+            lua_pushstring(L, result->s_name);
+            return 1;
+        }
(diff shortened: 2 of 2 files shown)


agraef:
Looks good to me. You asked for comments from @timothyschoen in #52 and he didn't object, so I'd say that this can go in now.

agraef:
Ben, maybe you can follow this up with some proper documentation of how set_args() works in graphics.txt?

agraef:
Just tested it with your provided test patch in vanilla, works perfectly AFAICT. So I'll just go head and release this as 0.12.17.

ben-wes:
> Looks good to me. You asked for comments from @timothyschoen in #52 and he didn't object, so I'd say that this can go in now.

cool, thanks a lot for merging and all the comments in #52 and here! i assume that the principle functionality is as expected then - so even if there might be objections about implementation details at some point, it's hopefully not critical!

> [...] maybe you can follow this up [...]

sure, i'm on it! will create a PR for that later today.

... and great to see all those updates to purr data btw!

agraef:
If I had any objections about the implementation, I'd have voiced them here. No, you implemented everything exactly as I would have done it myself.

> sure, i'm on it! will create a PR for that later today.

Great, looking forward to it!

... and great to see all those updates to purr data btw!

Oh boy, that was a lot of work this time, but I think that I fixed every single bug and usability issue that my students and I came across during the last 6 semesters. Almost 200 commits went into this release, and except for maybe a dozen commits from the GSoC 2024 contributions, I did all those myself since the end of June. I'm tired now. :)

PR #57, opened 2024-09-27
Multiple paint layers

timothyschoen:
Draft for multi-layer painting for pdlua.

Currently, it works like this:

Apart from your normal paint function, you can define extra draw callbacks that draw on top of it:

```
function hello:paint(g)
    g:set_color(50, 50, 50)
    g:fill_all()
end

function hello:paint_layer_2(g)
    g:set_color(250, 84, 108, 1)
    g:fill_ellipse(self.circle_x, self.circle_y, self.circle_radius, self.circle_radius)
    pd.post("layer 2")
end

function hello:paint_layer_3(g)
    -- Draggable rectangle
    g:set_color(66, 207, 201, 1)
    g:fill_rounded_rect(self.draggable_rect_x, self.draggable_rect_y, self.draggable_rect_size, self.draggable_rect_size, 5)
    g:set_color(0, 0, 0, 1)
    g:draw_text("Drag\n me!", self.draggable_rect_x + 8, self.draggable_rect_y + 10, self.draggable_rect_size, 12)
end
```

Then, you can repaint the layers separately by calling

```
repaint()   -- Repaints all layers
repaint(0) -- Also repaints all layers
repaint(1)  -- Repaints only your regular paint(g) function
repaint(2)  -- Repaints only the paint_layer_2 function
repaint(N)  -- Repaints only the paint_layer_N function
```

The primary motivation is performance. If you have a static background with moving foreground, it won't have to repaint everything all the time.

Currently implemented for pd-vanilla and plugdata, so purr-data still needs to be done. I might even have broken the purr-data implementation a bit. So far it seems to work well, but the pd-vanilla version could use some more testing.


timothyschoen:
[hello-gui.txt](https://github.com/user-attachments/files/17157287/hello-gui.txt)


There's the hello-gui example with multi-layer drawing. Unfortunately, github doesn't accept .pd_lua files

timothyschoen:
Screenshot 2024-09-27 at 03 56 31

Btw, after merging in the latest changes, I get this error in plugdata. Any idea why?

Seems like the code is only passing one string to _setrequirepath while the lua code expects 2 arguments, or maybe I messed something up while merging.

ben-wes:
cool! tested this with vanilla here - works perfectly well. can't respond to the `_setrequirepath` issue in plugdata though.

agraef:
> Seems like the code is only passing one string to _setrequirepath while the lua code expects 2 arguments, or maybe I messed something up while merging.

I remember that I added the extra pdlua_path argument at some point in 0.12, but that's long gone again. So it looks like your pd-lua source is in an inconsistent state, and you're running an outdated pd.lua there. There have been lots of changes in the entire source since you last merged, so make sure that you get all of it. :)

agraef:
Ok, this will need some time so that I get the new gui interface ported to Purr Data, but I'll do it asap, hopefully before the semester starts.

But I have two immediate remarks:

- The latest source already has a declaration for sys_trytoopenone in pdlua.c, please remove the one in your changeset.

- You renamed the `class` member in  `t_pdlua` to `pdlua_class`, is there an important reason for that? As you might have noticed, I have a second "shadow" class for the gfx widget behavior now so that I can swap them out as needed. That member is still named `class_gfx`, so the naming is inconsistent now.

agraef:
Also, what about the repaint method, can it still be invoked without arguments as before? Otherwise it would break a lot of existing Lua objects.

agraef:
Other than that, the changeset looks good to me, I just need a little time for the purr-data port.

agraef:
Also, Ben, thanks for testing the vanilla variant, it's good to know that this works, since it means that all I'll have to do is to massage the JavaScript code!

timothyschoen:
> The latest source already has a declaration for sys_trytoopenone in pdlua.c, can you please remove the one in your changeset? Or maybe change the existing one so that it works for you?

I'll do that! Probably put that in at some point to get it to compile, before merging your changes.

>  You renamed the `class` member in  `t_pdlua` to `pdlua_class`, is there an important reason for that? As you might have noticed, I have a second "shadow" class for the gfx widget behavior now so that I can swap them out as needed. That member is still named `class_gfx`, so the naming is inconsistent now.

Yeah, the plugdata-side implementation for pdlua imports the "pdlua.h" file, and "class" is a reserved keyword in C++, so it wouldn't compile. By changing the to "pdlua_class", the header is valid C and C++ code. Now I'm writing this though, I realise that I should have done "extern 'C'" instead, so I guess we could revert it.

ben-wes:
just quickly adding to this: Pd crashes with some of my graphical objects now. i'll investigate more later what the relevant difference there is between the gui example and my objects!

agraef:
> Now I'm writing this though, I realise that I should have done "extern 'C'" instead, so I guess we could revert it.

If it works, please do. Otherwise the name of the shadow class needs to be changed accordingly. I despise inconsistent naming, it makes code harder to read.

> Pd crashes with some of my graphical objects now.

That's not good, please report it here so that Tim can fix it. Pd-Lua gfx code should **never** crash Pd, no matter how broken it is.

ben-wes:
@timothyschoen one finding concerning the crash: it's obviously connected to a clocked `repaint()` or `repaint(0)` method. your new example above works for me. but the old one also crashes here (mac m2).

ben-wes:
> Also, what about the repaint method, can it still be invoked without arguments as before?

yep - tim mentioned that above: `repaint()   -- Repaints all layers`

timothyschoen:
> > Now I'm writing this though, I realise that I should have done "extern 'C'" instead, so I guess we could revert it.
> 
> If it works, please do. Otherwise the name of the shadow class needs to be changed accordingly. I despise inconsistent naming, it makes code harder to read.

Unfortunately, it didn't work. Bascially any name that isn't "class" would be okay for me.

> > Pd crashes with some of my graphical objects now.
> 
> That's not good, please report it here so that Tim can fix it. Pd-Lua gfx code should **never** crash Pd, no matter how broken it is.

> @timothyschoen one finding concerning the crash: it's obviously connected to a clocked `repaint()` or `repaint(0)` method. your new example above works for me. but the old one also crashes here (mac m2).

Gonna check now why that is!

timothyschoen:
> > Seems like the code is only passing one string to _setrequirepath while the lua code expects 2 arguments, or maybe I messed something up while merging.
> 
> I remember that I added the extra pdlua_path argument at some point in 0.12, but that's long gone again. So it looks like your pd-lua source is in an inconsistent state, and you're running an outdated pd.lua there. There have been lots of changes in the entire source since you last merged, so make sure that you get all of it. :)

Ah, that's probably it. Thanks!

timothyschoen:
I think f8e94a3 fixes the crash. The indexing for layers was a bit messy, since I kept the 1-indexing from lua in the C code, which also made it allocate space for one more layer than we need. This messy logic led me to make a mistake that caused the crash. Should be all good now

ben-wes:
cool, thanks! for some reason, i now get errors when redrawing the first or all layers in a little test patch:

~~~
"uplevel #0 $docmds"(Tcl) UNHANDLED ERROR: tagOrId ".l115c9071c0" doesn't match any items
    while executing
".x600000f80600.c lower {.l015c9071c0} {.l115c9071c0} "
    ("uplevel" body line 6)
    invoked from within
"uplevel #0 $docmds"
~~~

... your new hello-gui.pd_lua looks good though and also the previous version works again. i currently don't see what i'm doing wrong. just using this test:

~~~lua
local t = pd.Class:new():register("gui-test")

function t:initialize(sel, atoms)
    self.inlets = 1
    self.delay_time = 50
    self:set_size(200, 200)
    return true
end

function t:postinitialize()
    self.clock = pd.Clock:new():register(self, "tick")
    self.clock:delay(self.delay_time)
end

function t:finalize()
    self.clock:destruct()
end

function t:paint(g)
    local b = math.random()*255
    g:set_color(b, b, b);
    g:fill_all();
end

function t:paint_layer_2(g)
    local b = math.random()*255
    g:set_color(b, b, b);
    g:fill_rect(1, 1, 99, 99);
end

function t:tick()
    self:repaint(1)
    self.clock:delay(self.delay_time)
end
~~~

ben-wes:
ah, i see now that a `bang` to `[hello_gui]` also gives me this error - so it's obviously connected to redrawing the first layer via `repaint()`, `repaint(0)` or `repaint(1)`.

@timothyschoen in this context: i guess, the syntax for the layers is intuitive and analogous to the inlet messages - i'm just wondering if `paint_layer_1(g)` should also be allowed synonymous to just `paint(g)`?

timothyschoen:
> ah, i see now that a `bang` to `[hello_gui]` also gives me this error - so it's obviously connected to redrawing the first layer via `repaint()`, `repaint(0)` or `repaint(1)`.
> 
> @timothyschoen in this context: i guess, the syntax for the layers is intuitive and analogous to the inlet messages - i'm just wondering if `paint_layer_1(g)` should also be allowed synonymous to just `paint(g)`?

Ah, that makes sense, since I did change the logic for that. Must've made a mistake somewhere. I'll take look, should be fixable!

agraef:
> Unfortunately, it didn't work. Bascially any name that isn't "class" would be okay for me.

Yeah, that's because it's a reserved keyword in C++. Then keep it the way it is, but please also change `class_gfx` to `pdlua_class_gfx`.

Looking at `pd.Class:repaint` in pd.lua, it looks like the indentation is messed up. Also, there's a lot of repeated or similar code in there now, maybe the entire method could be refactored so that it becomes a little tidier again?

OK, I'll wait until you fixed the teething problems before I have a go at porting all the gui changes.

**Note to self:**
- Also need to check that pdx.lua still works.
- Also need to update the documentation (doc/graphics.txt, pdlua-help.pd -- graphics subpatch) and examples (pdlua/hello-gui.pd_lua, see Tim's attachment above).
- Check which of the other gui examples need fixing up to make them work better, without the frame rate throttling kludges that I added earlier if possible.

(Ben, you could maybe give me a helping hand with any of these items if you have the time.)

timothyschoen:
Should be fixed!

ben-wes:
> Should be fixed!

can confirm that the warning is gone and it works now for a single gui object instance!

if i load a patch with 2 instances, one is somehow "stealing" the first layer though. this also happens when i'm redrawing the first layer of one: it's disappearing for the other object then.

cutting one of the objects then also gives me:
~~~
(Tcl) UNHANDLED ERROR: tagOrId "(null)" doesn't match any items
    while executing
".x600000a9c300.c raise {.l1150c06fb0} {(null)} "
    ("uplevel" body line 6)
    invoked from within
"uplevel #0 $docmds"
~~~

ben-wes:
damn - i realize now that it's a bit more complicated: it actually all looks fine when the clocked repainting isn't started immediately. so it seems like a timing issue between the attempt to repaint and the first paint maybe?

ben-wes:
i'm wondering now if it might be an option to delay the postinitialize() after the first paint? would that allow to manage clocked repainting without the mentioned risk? ... and would it possibly also allow to read the arguments via get_args() at that point then? - admittedly, the latter is off-topic here.

timothyschoen:
> if i load a patch with 2 instances, one is somehow "stealing" the first layer though. this also happens when i'm redrawing the first layer of one: it's disappearing for the other object then.
> 
> cutting one of the objects then also gives me:
> 
> ```
> (Tcl) UNHANDLED ERROR: tagOrId "(null)" doesn't match any items
>     while executing
> ".x600000a9c300.c raise {.l1150c06fb0} {(null)} "
>     ("uplevel" body line 6)
>     invoked from within
> "uplevel #0 $docmds"
> ```

I can't reproduce this one yet. You mean just instantiate the object multiple times right, like this:

Screenshot 2024-09-30 at 14 09 41

Cutting didn't give me any error either. Maybe I'm missing something?

ben-wes:
i'm only testing vanilla here so far, i admit. with [this test patch](https://github.com/user-attachments/files/17190711/test-lua-gui.zip), i see 3 cases (when redrawing via a tick method that's immediately triggered through a clock from `postinitialize` - see comment in pd_lua):
* for `repaint()`, i see an empty canvas which is not responsive anymore (i need to kill Pd then)
* for `repaint(1)`, everything seems to work
* for `repaint(2)`, i get the situation described above, where the 1st layer is only drawn for 1 of the instances

timothyschoen:
Thanks for clarifying, I was able to reproduce both issues. And I think they are both fixed now!

timothyschoen:
> cool, thanks a lot! that fixes it all for me here!
> 
> if you want, you can add these 2 commits to the PR - using your new hello-gui (which is really a lot faster than before that way) and consistent naming for `pdlua_class_gfx`:
> 
> [timothyschoen/pd-lua@master...ben-wes:pd-lua:gfx/minor-changes](https://github.com/timothyschoen/pd-lua/compare/master...ben-wes:pd-lua:gfx/minor-changes)

Done!

ben-wes:
> (Ben, you could maybe give me a helping hand with any of these items if you have the time.)

@agraef : i'll gladly do more testing and look into updating examples and help! i'm considering to handle the latter in a separate issue and PR though since this one is getting a bit long and imho can be merged if no more issues occur (will do more vanilla testing now and report back).

ben-wes:
@timothyschoen i tested a bit more now and it mostly works really well! - for some reason, i still experience glitches and also cases where drawing heavily slows down after some time. i assume that these might be connected ... somehow, elements get confused here. this happened for example after i had a few instances of hello-gui on the canvas, deleted them and created an instance of my dualsense object:

image

... just wondering now how to reliably reproduce this.

besides that, i have a branch here with 2 more little changes for the help and graphics.txt:
https://github.com/timothyschoen/pd-lua/compare/master...ben-wes:pd-lua:doc/multilayers

ben-wes:
a simple way to reproduce these glitches with the hello-gui object is to just create one on a new canvas and then cut/paste until these artifacts appear. that works for me every time here while testing now.

for testing the updated version with multi-layered drawing of the dualsense object [dsshow], you'd need these 2 files:

* https://github.com/ben-wes/pd-dualsense/blob/gui/multilayer/dsshow-shapes.lua
* https://github.com/ben-wes/pd-dualsense/blob/gui/multilayer/dsshow.pd_lua

... with [dsshow], the drawing also slows down quite quickly. AFAICT, the main difference between the 2 objects is that [dsshow] also repaints the 1st layer (besides the 3rd one). my assumption here would then be that the slowdown is connected to repainting the 1st layer then ...

ben-wes:
@timothyschoen adding this video here to illustrate how i manage to reproduce these glitches:


https://github.com/user-attachments/assets/711e46d7-7ea3-4103-9b52-b56d4ac8cf54

timothyschoen:
Looks like the only elements that stroke a path. So must be a mistake there!

timothyschoen:
Fixed that! I'm not sure this also solves your slowdown issue though?

ben-wes:
> Fixed that! I'm not sure this also solves your slowdown issue though?

great! works flawlessly now. i don't manage to make anything glitch or crash. and it absolutely solves the slowdown issue as well! :)

agraef:
So is everything working fine in vanilla now? I'm asking because this PR is still marked as Draft/WIP.

agraef:
> i'll gladly do more testing and look into updating examples and help! i'm considering to handle the latter in a separate issue and PR though since this one is getting a bit long and imho can be merged if no more issues occur (will do more vanilla testing now and report back).

Thanks.Ben, for the offer and all the testing that you already did! Yes, I agree that the examples can be added in a later stage, when I implemented the new api in Purr Data.

agraef:
Tim, the extra declaration of sys_trytoopenone is still in the changeset, can you please remove it?

ben-wes:
> Tim, the extra declaration of sys_trytoopenone is still in the changeset, can you please remove it?

@timothyschoen if i see it correctly, this PR on your side would fix that?
https://github.com/timothyschoen/pd-lua/compare/master...alcomposer:pd-lua:fix-msvc-compile

ben-wes:
> So is everything working fine in vanilla now? I'm asking because this PR is still marked as Draft/WIP.

based on my vanilla tests, this looks ready! (and i see that tim also marked it as ready for review). the only question that was still open for me from above is this:

> i'm [...] wondering if `paint_layer_1()` should also be allowed synonymous to just `paint()`?

... but this is a minor detail and not critical - and maybe the redundancy for the sake of consistency is also not a good solution (since we certainly can't omit `paint()`)?

timothyschoen:
> > So is everything working fine in vanilla now? I'm asking because this PR is still marked as Draft/WIP.
> 
> based on my vanilla tests, this looks ready! (and i see that tim also marked it as ready for review). the only question that was still open for me from above is this:
> 
> > i'm [...] wondering if `paint_layer_1()` should also be allowed synonymous to just `paint()`?
> 
> ... but this is a minor detail and not critical - and maybe the redundancy for the sake of consistency is also not a good solution (since we certainly can't omit `paint()`)?

We could do this, but it would add a bit of code duplication if we need to check for both paint(g) and paint_layer_1(g). Also, what would happen if you define both?

timothyschoen:
> > Tim, the extra declaration of sys_trytoopenone is still in the changeset, can you please remove it?
> 
> @timothyschoen if i see it correctly, this PR on your side would fix that? [timothyschoen/pd-lua@master...alcomposer:pd-lua:fix-msvc-compile](https://github.com/timothyschoen/pd-lua/compare/master...alcomposer:pd-lua:fix-msvc-compile)

Yep you're right! Merged that instead, since if also fixes MSVC compilation

ben-wes:
> We could do this, but it would add a bit of code duplication if we need to check for both paint(g) and paint_layer_1(g). Also, what would happen if you define both?

yeah, the duplication is a bit ugly. and the last point is a very good question. let's skip this and properly document it (i'll do that)! :)

timothyschoen:
Hold off from merging for now, something broke in plugdata, I'd like to fix that too

timothyschoen:
OK that was easier than I thought, should be all good again

timothyschoen:
I had to use the outcommented line below this:

https://github.com/agraef/pd-lua/blob/07200b079ab5c61e18c36eaed5db9f0facf88fa8/pdlua.c#L2999

Otherwise, it will not set up the loader in plugdata. The return code I get is 2, so you can see how that would make it not set up the loader. It's unclear to me why this is the case, but using the line that was already there but outcommented fixes it.

I need to verify this, but I have a hunch that it has something to do with the stricter Documents folder permissions on macOS Sequoia (which is what I'm testing on).

agraef:
> I had to use the outcommented line below this:

That's wrong, I was just going to complain about it. Please revert. That line should actually be *removed*, not uncommented.

Note that the call to lua_load() already happens a few lines above that line, so you're just repeating that call again now (using some stale code which might actually be the wrong call, depending on the Lua library version -- remember that pd-lua is being used in a lot of different environments).

Whatever the problem is that you have there in plugdata, you need to solve it another way, without borking the pd-lua source. :roll_eyes:

agraef:
Note that just adding that call there also breaks the entire call logic in that part of pdlua_setup(): first load the Lua file, then execute that loaded chunk and check the result value. What you're doing now, is load the Lua file, then execute that loaded chunk and *not* check its result, then just load the Lua file *again*, leaving the stack in an incorrect state. That's just totally broken, nothing will work.

agraef:
Yes, I know that pdlua_setup() is complicated, and it's getting late. But this is one of the most important pieces of pd-lua, and should be treated with utmost respect. :smiley: Only touch it if you know *exactly* what you're doing and understand the Lua stack content at every single line of the routine.

agraef:
To make sure that this doesn't happen again, I just removed that commented bogus code line while fixing another, unrelated bug, that's why your PR has a merge conflict now. Reverting your change should fix that.

timothyschoen:
Done! I'll see why it didn't work here...

agraef:
Thanks! I guess that it's my turn now...

timothyschoen:
Ah, the problem is that I didn't copy pdx.lua into the search path, so I can fix that on my side. Without that, the lua_pcall function returns a result value of 2, since it can't find the include file.

I'd say that if pdlua doesn't start without this file, we should put it in the top level directory of the repo instead of in tutorial/examples/pdx.lua? It's not just an example, it's a dependency without which pdlua doesn't initialise correctly.

agraef:
> I'd say that if pdlua doesn't start without this file, we should put it in the top level directory of the repo

You're right. Inertia and laziness prevented that so far. :) There's also some other files that should go there, have a look at the definition of pdx_files in the toplevel pd-lua Makefile. (At least pd-remote.pd will be needed for the live-coding support.)

agraef:
Ok, I pulled your branch into my working copy, and started a deep dive into the source now, in order to figure out what exactly I need to change in the PURR_DATA part. There's still some stuff that I need to understand a little better, it would be nice if you could help me out there. (Sorry if you already provided that information, but this thread is getting a bit long, and I don't have the time to dig through it again.)

- As I understand it, layer 0 (and thus layer_tags[0]) is never used, right? But shouldn't layer_tags[0] at least be initialized to *something*, if only NULL? So that you get a defined value in case of an off-by-one bug?
- In start_paint() at L838, you retrieve the layer number from argument 2. I think that this will give a Lua exception if that argument isn't present, which is good. But what happens if the argument is zero or negative? You'd then access layer_tags[0] or, worse, layer_tags[i] with i<0. Are you really *absolutely* 100% sure that this can never happen? If so, why? I'd say that a little bit of defensive programming can't hurt there, so it would be prudent to check that layer is > 0 and bail out if it's not, to prevent crashing Pd.
- At L841-846 you're creating new layer tags as needed. But you're only initializing the current one. What happens if layer != gfx->num_layers? Again, are you really *absolutely* 100% sure that this can never happen? If so, why? Imagine that the user defines paint_layer_2 and paint_layer_4, but not paint_layer_3 at first? But then realizes his mistake and defines paint_layer_3 afterwards and reloads the script using live-coding. In that case layer_tags[3] will still be uninitialized which will crash Pd or at least cause some damage somewhere else because of the uninitialized char*.

I think that it's for now, but I'll let you know if I run into other problems.

agraef:
Note that by "really *absolutely* 100% sure" I mean "could produce a mathematical proof of the fact if I needed to". In all other cases defensive programming is in order! :smirk: Better safe than sorry, IMHO.

agraef:
Another question: Where is the memory of the layer tags (both the char * pointers and the array itself) freed again? I don't see anything in the code.

timothyschoen:
> Ok, I pulled your branch into my working copy, and started a deep dive into the source now, in order to figure out what exactly I need to change in the PURR_DATA part. There's still some stuff that I need to understand a little better, it would be nice if you could help me out there. (Sorry if you already provided that information, but this thread is getting a bit long, and I don't have the time to dig through it again.)

No problem!

>  As I understand it, layer 0 (and thus layer_tags[0]) is never used, right? But shouldn't layer_tags[0] at least be initialized to _something_, if only NULL? So that you get a defined value in case of an off-by-one bug?

Layer tag 0 should get initialised and used. It's associated with the regular paint(g) call. It's at least used when we call pdlua_gfx_clear, to clear the first layer when it needs to be updated. It also gets used in end_paint to do ordering. Note that we get the layer index as `luaL_checknumber(L, 2) - 1`, so we change it to be 0-indexed.


>  In start_paint() at L838, you retrieve the layer number from argument 2. I think that this will give a Lua exception if that argument isn't present, which is good. But what happens if the argument is zero or negative? You'd then access layer_tags[0] or, worse, layer_tags[i] with i<0. Are you really _absolutely_ 100% sure that this can never happen? If so, why? I'd say that a little bit of defensive programming can't hurt there, so it would be prudent to check that layer is > 0 and bail out if it's not, to prevent crashing Pd.

The call to pdlua_start_paint is made in pd.lua, so we can control the values that go in there. So it shouldn't be able to go wrong. If you want to add an extra check in case the user incorrectly modifies pd.lua, that's fine with me.

>  At L841-846 you're creating new layer tags as needed. But you're only initializing the current one. What happens if layer != gfx->num_layers? Again, are you really _absolutely_ 100% sure that this can never happen? If so, why? Imagine that the user defines paint_layer_2 and paint_layer_4, but not paint_layer_3 at first? But then realizes his mistake and defines paint_layer_3 afterwards and reloads the script using live-coding. In that case layer_tags[3] will still be uninitialized which will crash Pd or at least cause some damage somewhere else because of the uninitialized char*.

I see what you mean, I thought I made sure that can't happen but looks like it can happen! So we need to fix that. Leaving holes is not allowed currently, but it also shouldn't crash because of it. I've just made a change to pd.lua that should make sure this can't happen (and also makes it easier to read), but if you want to be extra sure, we can also update the C code to make sure it won't crash. I think it should never happen, the safer the better.

timothyschoen:
> Another question: Where is the memory of the layer tags (both the char * pointers and the array itself) freed again? I don't see anything in the code.

Oops! I still need to do that

timothyschoen:
I fixed the memory leaks, hope it's not too much of a hassle for you to merge in

timothyschoen:
Random thought, but perhaps we should heap allocate the pdlua_gfx object, so that it doesn't take up memory for non-gui objects?

agraef:
No worries, my changes are confined to the PURR_DATA parts (at least so far), so I can just rebase on your latest and be done with it.

agraef:
> I've just made a change to pd.lua that should make sure this can't happen

Ok, I see what you did there in https://github.com/agraef/pd-lua/pull/57/commits/696d227a20671ff99020c6e266dcac8849a72778, but it's not helping, sorry. :) Let me elaborate.

Your proof of correctness rests on the assumption that pd.Class:repaint(0) is called before anything else so that the entire layer tags array gets populated there, and that's initially true. But that assumption breaks down with live-coding, just as it did before, and your commit doesn't help with this.

Consider this use case: The user starts out with a simple animation with just one layer, layer 1. The object is created with this layer, at which point pd.Class:repaint(0) gets called and the tags table is populated with one entry. (Or maybe 2? I still don't quite get your numbering scheme there, I'm afraid.)

Then the user adds two extra layers, layer 2 and 3, and adds messages to kick off those parts of the animation using separate clocks. He reloads the script, which will *not* cause pd.Class:repaint(0)  to be called again, because the object is already instantiated and running. You can *not* prevent this from happening, because the raison d'être of live-coding is precisely that the underlying Lua code can be swapped out *at any time* without any disruptive state changes in already running instances of the object.

While the animation keeps running, subsequent pd.Class:repaint() calls will pick up the new layers 2 and 3 and, as they are consecutive, will add them both to the paint_funcs table. But their entries in the layer tags array on the C side still haven't been created yet, because pd.Class:repaint(0) didn't run (and never will again in this example).

But now the user, let's call him Albert D'Evil, adds the requisite messages for the layer 2 and 3 animations to his patch and decides that he wants to kick off the layer 3 animation first, because ... why not? :)) So pd.Class:repaint(3) is called and there you got your hole at index 2 in the tag table now. As soon as the user then kicks off the layer 2 animation, Pd will crash and burn.

You see, with live coding the whole environment becomes a whole lot more dynamic, so all your assumptions about the execution sequence go out of the window, no matter how hard you try. The *only* way to fix this problem is to cope with the out-of-order layer initializations on the C side when populating the tags table. And it's *very* easy to do, just do the snprintf thing right there in in start_paint() with *all* newly created layer tags instead of just a single one.

I also don't like that you're now constructing and checking the entire table of layers each time pd.Class:repaint() is called, even if it's just to update a single layer. That might become a performance hog if you have many layers, because in an animation that method might be called many times per second, and it will then have to reconstruct the entire layer functions table each time. So it kind of defeats the purpose of having layers which can be updated independently in the first place.

TL;DR: As https://github.com/agraef/pd-lua/pull/57/commits/696d227a20671ff99020c6e266dcac8849a72778 doesn't really help with the out-of-order layer tag creation problem, and it introduces substantial overhead for no real gain, I'd revert it, and implement the proper incremental initialization of the tags table in start_paint() that I sketched out above instead. Problem solved. :)

Of course, I can do it for you if I didn't explain it clearly enough.

agraef:
One more thing: You're using resizebytes() to resize the layer tags array in start_paint(), but malloc() for the initial allocation, and free() for deallocation in pdlua_gfx_free(). Which might be inconsequential if Pd's underlying implementation in fact uses malloc/realloc/free. But in any case it's inconsistent -- please use either malloc/realloc/free or their Pd equivalents, I think that's safer.

timothyschoen:
> which will _not_ cause pd.Class:repaint(0) to be called again

Ah, I hadn't considered that. Makes sense, gonna work in it! I'll also revert the pd.lua change.

timothyschoen:
The layer indices in Lua should be:

0: Not a real layer, but will trigger a repaint of all layers
1: Regular paint(g) function call
2: paint_layer_2(g) function call
n: paint_layer_n(g) function call

When we convert that into C, we subtract 1 so that we can use it as an index for the layers array, so:
-1: not a real layer, full repaint
0: repaint(g)
etc.

Hope that helps, let me know if you spot anywhere I did it wrong.

timothyschoen:
That last problem is kind of a fundamental one. We can't render on top of layer 2 if it doesn't exist. I see 2 possible solutions:

Solution 1: We create an invisible dummy tag for skipped layers, so we can order things correctly. We'll need to correctly position the dummy layer, and after that the real layer relative to that. IMO: this feels bad because now we are in a state where layer 3 is visible but layer 2 is not. I don't feel like that should be possible.

Solution 2: If we notice a hole in the paint function chain, we trigger a full repaint. We consider it an invalid state, and solve it by re-rendering everything. I've verified that this works perfectly and solves all our problems, but it feels pretty nasty code-wise, because we're synchronously repainting all the layers, inside a layer repaint call:

```
        int layer = luaL_checknumber(L, 2) - 1;
        if(layer > gfx->num_layers) // If we get here, we have skipped a layer. This isn't allowed, so we should instead repaint everything
        {
            pdlua_gfx_repaint(obj, 0);
            lua_pushnil(L);
            return 1;
        }
        else if(layer >= gfx->num_layers)
        {
            int new_num_layers = layer + 1;
            if(gfx->layer_tags)
                gfx->layer_tags = resizebytes(gfx->layer_tags, sizeof(char*) * gfx->num_layers, sizeof(char*) * new_num_layers);
            else
                gfx->layer_tags = getbytes(sizeof(char*));
            
            gfx->layer_tags[layer] = getbytes(64);
            snprintf(gfx->layer_tags[layer], 64, ".l%i%lx", layer, (long)obj);
            gfx->num_layers = new_num_layers;
        }
``` 

For now, I'm going with solution 2, because I feel like this actually solves the problem instead of trying to make an invalid state work. But feel free to work on this if you have the time.

agraef:
> Hope that helps, let me know if you spot anywhere I did it wrong.

No, it makes perfect sense now, thanks for explaining it once more and sorry for my slow-wittedness. :)

> It will try to position layer 3 on top of layer 2, but tcl/tk is not aware layer 2 exists yet, so it will print an error. So we need to find a way around that too.

Right, I hadn't thought about this bit, because that `raise` command doesn't exist in the PURR_DATA code. I can do without it, because I'm just replacing the inner html in the DOM when repainting the object, which maintains the correct stacking order without any further ado.

Of course, with the layer implementation, I will now have to deal with that kind of situation in the PURR_DATA code, too. But I can actually solve this by simply adding the SVG equivalent of what you called a skipped layer (solution 1) right there in start_paint(). In HTML/SVG this is just an empty SVG group which may be filled with graphics objects later if and when the layer actually gets painted. This makes it possible to maintain the correct stacking order *without* having to resort to a full repaint.

It would be awesome if something like this could be implemented in vanilla's Tcl/Tk GUI as well, but I can see why you're running into toolkit limitations with this. :( While I still like Tcl/Tk for its simplicity, it shows its age there. Having the DOM and HTML+SVG rendering engine of a modern HTML browser at your fingertips makes all this much easier.

Ok, I think that I can continue with the purr-data port now. I haven't got very far with it yet, but I think that I have a good grasp of the layer architecture now, so this should be easy to do.

pdlua_gfx.h
```diff
@@ -915,6 +923,7 @@ static int end_paint(lua_State* L) {
     int xpos = text_xpix((t_object*)obj, obj->canvas);
     int ypos = text_ypix((t_object*)obj, obj->canvas);
 
+    // TODO: I don't think we need to call drawiofor on each layer?
```
agraef, line 926: I think that as long as the paint method can change the size of the object no matter what the layer is, we need to keep this for all layers, I'm afraid. However, we might be able to optimize this by recording the size in start_paint and then check in end_paint whether the size was changed.

pd.lua
```diff
@@ -462,41 +462,44 @@ function pd.Class:canvas_realizedollar(s)
...
+    local paint_layer_method = "paint_layer_" .. tostring(layer)
+        if type(self[paint_layer_method]) == "function" then
+            local g = _gfx_internal.start_paint(self._object, layer)
+             if g ~= nil then
+                self[paint_layer_method](self, g)
+                _gfx_internal.end_paint(g, layer)
+            end
         end
-   end
+     end
```
agraef, line 502: The indentation got botched there with the revert, but I can fix that in post.

pdlua_gfx.h
```diff
@@ -968,14 +1023,14 @@ static int stroke_ellipse(lua_State* L) {
...
+    const char* tags[] = { gfx->object_tag, register_drawing(gfx), gfx->current_layer_tag };
 
 #ifndef PURR_DATA
-    pdgui_vmess(0, "crr iiii ri rs rS", cnv, "create", "oval", x1, y1, x2, y2, "-width", line_width, "-outline", gfx->current_color, "-tags", 2, tags);
+    pdgui_vmess(0, "crr iiii ri rs rS", cnv, "create", "oval", x1, y1, x2, y2, "-width", line_width, "-outline", gfx->current_color, "-tags", 3, tags);
 #else // PURR_DATA
     int x0 = text_xpix((t_object*)obj, obj->canvas);
     int y0 = text_ypix((t_object*)obj, obj->canvas);
-    gui_vmess("gui_luagfx_stroke_ellipse", "xsssiiiii", cnv, tags[0], tags[1],
+    gui_vmess("gui_luagfx_stroke_ellipse", "xssssiiiii", cnv, tags[0], tags[1],
```
agraef, line 1033: Note to self: Tim encroached on PURR_DATA terrain here, need to fix that in post.

agraef:
Ok, I think that this will do: https://github.com/agraef/pd-lua/compare/d0dd9a7221827fe79e155a6062009274e7daf0f6...496b965b751a24354c943ee1d365ec0ae4c22a8f

Great job, Tim, as always! Porting this for purr-data was surprisingly easy, congrats on the excellent design!

For the record: In the purr-data implementation, this just required adding the layer management in start_paint, and adjusting pdlua_gfx_clear accordingly. The graphics operations are virtually unchanged, except that they now take the layer tag in lieu of the object tag as argument. And in the purr-data pdgui.js module, I just needed to add a single new gui_luagfx_new_layer function which takes care of creating the layers (a.k.a. svg 'g' objects) which act as containers for the svg graphics objects now. And this should be entirely backward-compatible, just like it is in vanilla -- with older pd-lua versions, it will just continue to use the gobj as parent instead of the layers and never call gui_luagfx_new_layer at all. At least that's what I hope. :)

I still need to test this, which I'll do over the weekend. In particular, I need to make sure that the generated layer tags don't collide with tags we already use in the purr-data GUI, but that seems unlikely.

Alright, I'm going to merge this now!

pdlua_gfx.h
timothyschoen: The set_size method can be called from anywhere, not just the paint functions. So yeah, I think we need to keep track of it.

agraef:
All done and dusted, 0.12.19 is tagged and currently building.

I still had to fix a few glitches and updated the docs, so Tim, please make sure that you merge back the latest into your fork.

pdlua_gfx.h
agraef: Yeah, let's just keep this in mind for for future optimization opportunities. Rome wasn't build in a day either.

PR #60, opened 2024-10-07
Feature/multichannel

### LINKED ISSUE #53: multichannel support
and another discussion disguised as an issue. :)

---

it would be nice to have multichannel support for signals in pdlua. the (probably?) obvious way might be to have a table of tables in that case, where `#in` for `perform(in)` would tell the number of channels. but this is not backward compatible and possibly also not very efficient? ... although there could also be a new type like `MC_SIGNAL` for the inlets/outlets definitions.

another option might be to just have a flat table with all samples for all channels? the blocksize can be obtained through the `dsp()` function - so the number of channels could be derived in other ways (maybe also written somewhere from `dsp()` function?).

for outlets, the number of channels would need to be defined some way though, i guess ... not sure what the syntax there should be and what `perform()` should return.

ben-wes:
First attempt to close #53 that might cover all requirements.

Attaching some test files here once more:
[test-lua-mc.zip](https://github.com/user-attachments/files/17281982/test-lua-mc.zip)

Brief documentation:
* Lua dsp function passes inlet channel counts with additional table:
  ~~~lua
  function myobj:dsp(samplerate, blocksize, inchans)
  ~~~
* `signal_setmultiout` is exposed and can be used to set outlet channel counts in dsp function. example for setting 1st outlet to 2 channels:
  ~~~lua
  self:signal_setmultiout(1, 2)
  ~~~
* samples of additional channels are concatenated to samples of first channel for iolets. this would be one way to switch channel 1 and 2:
  ~~~lua
  function myobj:perform(in1)
    out1 = {}
    for i = 1, self.blocksize do
      out1[i] = in1[i + self.blocksize]
      out1[i + self.blocksize] = in1[i]
    end
    return out1
  end
  ~~~
* `PD_MULTICHANNEL` can be set to 0 to build without multichannel support (automatically gets set depending on Pd src otherwise)

EDIT: successfully tested:
* macos arm64
* windows amd64


pdlua.c
```diff
@@ -1057,22 +1070,31 @@ static t_int *pdlua_perform(t_int *w){
     for (int i = 0; i < o->siginlets; i++)
     {
         lua_newtable(__L());
-        t_float *in = (t_float*)w[i + 3];
-        for (int j = 0; j < nblock; j++)
+        t_signal *sig = (t_signal *)(w[2 + i]);
+        t_float *in = sig->s_vec;
+#if PD_HAVE_MULTICHANNEL
+        int nchans = sig->s_nchans;
```
ben-wes, line 1076: after looking at this once more, i thought this should also have a runtime check - so i'm currently a bit irritated why i get input samples in Pd0.53 with a version compiled with Pd0.55 ...

pdlua.c
```diff
@@ -1047,7 +1058,9 @@ static void pdlua_menu_open(t_pdlua *o)
 
 static t_int *pdlua_perform(t_int *w){
     t_pdlua *o = (t_pdlua *)(w[1]);
-    int nblock = (int)(w[2]);
+    t_signal *first_sig = (t_signal *)(w[2]);
+    int nblock = first_sig->s_n;
+    // FIXME: should we use a o->blocksize here that gets set in dsp?
```
ben-wes, line 1063: marking this comment i made in the code for actual review.

pdlua.c
```diff
@@ -1108,43 +1144,67 @@ static t_int *pdlua_perform(t_int *w){
...
-static void pdlua_dsp(t_pdlua *x, t_signal **sp){
+static void pdlua_dsp(t_pdlua *x, t_signal **sp) {
     int sum = x->siginlets + x->sigoutlets;
     if(sum == 0) return;
     x->sig_warned = 0;
-    
+
     PDLUA_DEBUG("pdlua_dsp: stack top %d", lua_gettop(__L()));
+
+    x->sp = sp; // FIXME: is this the way? (setting this for signal_setmultiout)
```
ben-wes, line 1161: i assume that this is fine since i don't see another simple way? ... the signal pointer is stored here only for `signal_setmultiout` to access it then.

ben-wes:
hold on - adding some more error management there ...

pdlua.c
```diff
@@ -2664,33 +2662,43 @@ static int pdlua_canvas_realizedollar(lua_State *L)
...
+        pd_error(NULL, "%s: signal_setmultiout: invalid arguments", src_info(L, msg));
+        return 0;
+    }
+
+    t_pdlua *x = (t_pdlua *)lua_touserdata(L, 1);
+    int outidx = lua_tointeger(L, 2) - 1;
+    int nchans = lua_tointeger(L, 3);
+    if (!(x && outidx >= 0 && outidx < x->sigoutlets))
+    {
+        pd_error(NULL, "%s: signal_setmultiout: must be called from dsp method", src_info(L, msg));
```
ben-wes, line 2678: not sure if it's valid to assume that this error is due to the wrong context.

ben-wes: i split these now in the following commit. still not sure about the assumption of the wrong context if x is not defined.

pdlua.c
ben-wes: not so irritated anymore. there was no sample table in that case. fixed with ce8a92dcf6a4458f1014a6344b85580ca78be451

Makefile
```diff
@@ -40,6 +41,7 @@ endef
 endif
 
 cflags = ${luaflags} -DPDLUA_VERSION="$(pdlua_version)"
+cflags += -DPD_MULTICHANNEL=$(PD_MULTICHANNEL)
```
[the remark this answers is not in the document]
ben-wes: should be solved with b51c9c0 (possibly not very elegant though)

pdlua.c
```diff
@@ -1088,13 +1112,27 @@ static t_int *pdlua_perform(t_int *w){
...
+
     for (int i = o->sigoutlets - 1; i >= 0; i--)
     {
-        t_float *out = (t_float*)w[i + 3 + o->siginlets];
-        for (int j = 0; j < nblock; j++)
+        if (!lua_istable(__L(), -1)) {
+            pd_error(o, "pdlua_perform: expected table for outlet %d", i);
+            lua_pop(__L(), 1);
+            continue;
+        }
```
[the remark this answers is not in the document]
ben-wes: remove with 5053034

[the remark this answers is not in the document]
agraef: I'm not sure that's correct. You're using data that's returned by user Lua code there, right? Then you *always* need to check that it's valid.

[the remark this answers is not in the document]
ben-wes: my assumption was that the checks here are enough (like in the current implementation by tim):
https://github.com/agraef/pd-lua/blob/505303451d4189457547760390e1a692c7574ece/pdlua.c#L1101-L1117 

... and there is also the fallback when writing the samples:
https://github.com/agraef/pd-lua/blob/505303451d4189457547760390e1a692c7574ece/pdlua.c#L1131-L1142

pdlua.c
ben-wes: closing this comment since this approach is probably not wrong.

pdlua.c
ben-wes: replaced this with `x->blocksize` now in https://github.com/agraef/pd-lua/commit/59632795544475e76de84dd1f32f7e870e470d57

pdlua.c
ben-wes: the error messages might still be improved - but they already give a good hint on what might go wrong. closing this comment for now.

ben-wes:
reverted a few formatting changes now to minimize the changeset.

pdlua.c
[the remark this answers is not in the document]
agraef: I don't understand. If I understood you correctly, you're saying that the `!lua_istable(__L(), -1)` check isn't needed anymore, and that this is because it's already done elsewhere? You obviously don't need to check exactly the same value twice, but...

This leaves me even more confused as before. ;-) Maybe we need to take a look at the entire code of `pdlua_perform` in context again to make sure that you didn't butcher it and all the required checks are still there?

agraef:
Are all the other design questions resolved now? Or did they just disappear because you force-pushed them away?

pdlua.c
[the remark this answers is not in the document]
agraef: Ok, I've pulled your branch now and read through the code, looks good to me!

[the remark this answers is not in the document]
agraef: In fact I was surprised at how deceptively simple it all looks. It still amazes that this will work as expected no matter what the host version is. Great job!

ben-wes:
> Are all the other design questions resolved now? Or did they just disappear because you force-pushed them away?

they are resolved imho, yes. i commented on them all here if you want to check the reasoning. i actually did that with additional commits instead of force-pushes so that the changes are visible.

> Sorry, seems that I'm a bit late so that you had to decide these issues yourself, but as I said I was busy with other things today...

no problem at all! i'm sorry if all my changes and comments put pressure on you - that was not the intention! that stuff was just very present in my thoughts and whenever i had some idea, i checked the code myself once more and gave it a try.

pdlua.c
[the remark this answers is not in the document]
ben-wes: cool - happy to hear! and the solution for handling different host versions is obviously not mine - thanks go to christof and his vstplugin here. :)

agraef:
Just compiled and ran it on Linux with purr-data, works without a hitch. If you're too exhausted to do the doc and example after this ordeal, that's ok, I can do it myself later today. :sweat_smile:

ben-wes:
> Ok, after reviewing the code I'm getting ready to merge this. Do you have a patch demonstrating the feature to go along with it? Something that looks a bit less scary than your test patch. ;-) This should go into pdlua/examples. Maybe a multichannel version of osci3d? (But please keep the old one for people still running old Pd versions without mc support.)

a multichannel version of osci3d sounds right - will do! for the simplest (?) lua mc object, this might do:

~~~lua
local test = pd.Class:new():register("test~")

function test:initialize(sel, atoms)
  self.inlets = {SIGNAL}
  self.outlets = {SIGNAL}
  return true
end

function test:dsp(samplerate, blocksize, inchans)
  pd.post(string.format("samplerate: %d, block size: %d, input channels: %s",
    samplerate, blocksize, inchans[1]))
  self.blocksize = blocksize
  self:signal_setmultiout(1, 2)
end

function test:perform(in1)
  out1 = {}
  for i = 1, self.blocksize do
    out1[i] = in1[i + self.blocksize]
    out1[i + self.blocksize] = in1[i]
  end
  return out1
end
~~~

> Also, can you still please write something about multi-channel in doc/graphics.txt and pdlua-help.pd / pd graphics?

sure! 9ed55f1058abe2002f3d550b61ff317278d37989

> Congrats on your first "real" code contribution! 🥇

thanks! 🎉 ... hope it will prove stable in the future, too. :)

> Now this _was_ substantial, but I hope that it wasn't quite as difficult as you expected, was it?

it was only possible with the help of all the existing code. but yeah - in the end, it looks surprisingly simple.

> When we meet next year, I really owe you a couple of beers now. :)

haha, i'm in!

agraef:
Yeah, let's keep it simple. Users should be able to understand that they need multi-channel support to run a  multi-channel Lua object. Same as with any kind of multi-channel external. You might add a caveat to your help patch, saying: :warning: "Grandpa, this is a MULTI-CHANNEL object, don't try to run this on your old and trusted Pd 0.32, it won't work!" They'll also notice that something is afoot if those snake~ objects won't instantiate. :)

Here's something I quickly cobbled together from your minimal test~ object from above. I think that'll do for now, so you can take your time with your mc examples.

test~

agraef:
Ok Ben, I take it that you're done testing in vanilla, and I did my own testing in purr-data, so I'm going to merge this now, and add the example I made. @timothyschoen You may want to give this a whirl yourself. I don't see why it should break in plugdata and libpd, except maybe for the dlsym hack used to detect whether the host supports multi-channel -- in that case we might have to add an `#ifdef PLUGDATA` or something similar to bypass that check at pdlua.c L3036 and just set the function pointer to the actual `signal_setmultiout` function.

agraef:
Well, as usual the patch looks rather crappy in vanilla, and the output~ abstraction there doesn't have any outputs so I had to rearrange things a bit. But anyway, here's that multichannel test patch again, now running in vanilla:

image

agraef:
Compare with this, then you might understand why I love purr-data so much:

image

agraef:
> i don't immediately see that test~ flips L<->R

Yes, it's hard to see, but you can hear it, especially when toying around with the parameters in the expr object.

> yeah ... crispy antialiasing-free! ;) ... luckily this is not an issue on macOS though.

Pd in macOS is ... tolerable, I'd say, but on Linux and Windows I just can't run vanilla for more than a few (low single digit) minutes or my eyes will start bleeding. (Quite literally-- my eyesight is not the best anymore, and those gui fonts are *much* too small for modern high-dpi displays, while the patch zoom has only two levels: too small and too large.) The vanilla devs really need to get their act together and throw out that Tcl/Tk GUI. It looks like something from the 1990s (which it is). Reminds me of Motif / Common Desktop Environment (CDE) which looked great at the time (at least compared to Athena), but not today. I still remember those days. :)

agraef:
Just for the record, and for testing purposes, here's that multichannel test patch: [multichannel.zip](https://github.com/user-attachments/files/17305256/multichannel.zip)

ben-wes:
> those gui fonts are _much_ too small for modern high-dpi displays, while the patch zoom has only two levels: too small and too large

i don't remember what i did there exactly - but when i was still on windows, i somehow managed to have an acceptable setup through display zoom settings etc. but sure - free zoom levels are great.

> The vanilla devs really need to get their act together and throw out that Tcl/Tk GUI. It looks like something from the 1990s (which it is).

i think there are some developments going on in that direction (starting a while ago and in the last days/weeks) and i'd be more than happy to see a different gui there as well! - mainly for performance reasons though. i admit that i still keep coming back to vanilla all the time for its visual simplicity. although design-wise, there are certainly some options to make patches more readable - and it's totally lacking a dark mode.

ben-wes:
oha! merged! thank you so much for all the feedback and support here!

agraef:
I only had to press a button :wink:  -- all kudos to you! Feel free to follow up with another PR when osci3d-wc is finished, or if you have anything else to add. Tim, the same goes for you if you discover any plugdata incompatibilities.

I'm currently working on adding that multi-channel note callout to the tutorial. After that, I think I'm just going to release 0.12.20 so that we get this out in the wild and people can start to test it.

agraef:
> and it's totally lacking a dark mode

Yeah, this is getting OT, but who cares... While we don't really have full gui themes like plugdata has (apart from the theming for Chromium which the OS desktop provides), there are canvas gui themes in purr-data. Here's "Inverted" in purr-data:

image

We also have a bunch of other, more colorful themes, including one for C64 fans. And for vanilla aficionados there is a theme as well (also an inverted version if you prefer dark mode):

image

But of course if you want a vanilla version with a dark mode, you can just use plugdata if you don't need all of purr-data's external libraries.

I don't really like to work in dark mode, so I usually prefer a light theme, and I think that purr-data's default theme actually looks pretty good and is easy on my aging eyes.

agraef:
Ben, as soon as you're done with earning your "Brötchen", can you please start work on updating the workflow so that it builds against a newer version of Pd (currently it's using Pd 0.52 on Linux, thus of course the mc support doesn't work there, even in the Pd 0.55-1 which I'm running here).

It's a good test that shows that your mc support magic actually works, but in this case I'd rather have a different result. :rofl:

ben-wes:
> as you're done with earning your "Brötchen", can you please start work on updating the workflow so that it builds against a newer version of Pd

weird - i have no idea about the mechanisms there, but obviously `sudo apt install puredata-dev` is not a good idea if the pd version didn't get updates anymore.

so before the Brötchen are done, i patched this with the github frontend here (and of course made a stupid mistake there first). but can you check these ubuntu builds here? https://github.com/ben-wes/pd-lua/actions/runs/11253875795

if they work, you could merge my master branch (and possibly squash these 2 commits into one).

agraef:
Got it, thanks for the quick fix!

PR #65, opened 2024-10-22
add optional alignment arg to draw_text

### LINKED ISSUE #64: Allow setting text alignment mode
currently, the text coordinate always is at the upper left of the text bounding box - obviously resulting in top-left aligned text.

For some use cases (especially when displaying numbers), right alignment would be a nice option to have.

For this, we could add a new alignment setting (similar to the way that colors are set) like `set_text_align(mode)`. An open question here is whether the vertical alignment is also necessary or if `left`, `center`, `right` are enough.

Tcl/Tk uses compass directions for this (see https://www.tcl.tk/man/tcl8.4/TkCmd/options.htm#M-anchor): `n, ne, e, se, s, sw, w, nw, center` ... that would be an option as well, but not sure if needed and if that would easily be adaptable for purr data and plugdata?

ben-wes:
closes #64

only handles the additional argument for vanilla.
opening this PR for further discussion since i'm not sure if that's "the way".

### DIFF
--- a/pd.lua
+++ b/pd.lua
@@ -628,6 +628,17 @@ DATA = 0
 SIGNAL = 1
 Colors = {background = 0, foreground = 1, outline = 2}
 
+-- Text alignment constants
+TOP_LEFT = 0
+TOP_CENTER = 1
+TOP_RIGHT = 2
+CENTER_LEFT = 3
+CENTER = 4
+CENTER_RIGHT = 5
+BOTTOM_LEFT = 6
+BOTTOM_CENTER = 7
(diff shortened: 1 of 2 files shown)


ben-wes:
added a minor change to achieve the handling for the optional argument with `luaL_optinteger`.

... while looking at this, i thought that we could similarly use this or `luaL_optnumber` in other cases for strokewidth and fontsize to make them optional?

EDIT: those additional optional arguments would then possibly look like this: https://github.com/agraef/pd-lua/commit/c429057211341c0ed39e4dcbdc869646271c8a20

PR #70, opened 2025-01-16
Add mouse_enter() and mouse_exit() for Pd vanilla

### LINKED ISSUE #61: additional `mouse_enter` and `mouse_exit` functions?
the `mouse_move(x, y)` callback can be used to create hover feedback in pdlua GUI objects.

unfortunately, this has one major restriction though: the position is only reported on changes inside the object box and these changes are not necessarily reporting continuous coordinates. so there's no way to reliably determine whether the mouse has left the object area (or "crossed a border").

one option to deal with this might be additional functions. i learnt that @timothyschoen had these implemented some time ago and then removed them again since they seemed slightly hacky for vanilla. maybe they could be reconsidered though since they would allow for some additional interaction patterns.

one question in this context is whether `mouse_enter` and `mouse_exit` could also be achieved easily with purr data?

ben-wes:
creating this PR mainly for discussion. tested with Pd vanilla

### questions
1. is the mechanism using parent canvas messages the correct way in Pd vanilla?
2. is the dragging behaviour expected that way (enter/exit are suppressed during drag, up and exit are reported outside on mouseup)?

### todos
* support plugdata and purr data

closes #61

### DIFF
--- a/pd.lua
+++ b/pd.lua
@@ -115,6 +115,12 @@ pd._mouseevent = function (object, x, y, event_type)
     if event_type == 3 and type(obj.mouse_drag) == "function" then
       obj:mouse_drag(x, y)
     end
+    if event_type == 4 and type(obj.mouse_enter) == "function" then
+      obj:mouse_enter(x, y)
+    end
+    if event_type == 5 and type(obj.mouse_exit) == "function" then
+      obj:mouse_exit(x, y)
+    end
   end
 end
 
--- a/pdlua.c
+++ b/pdlua.c
@@ -789,18 +789,27 @@ static void pdlua_motion(t_gobj *z, t_floatarg dx, t_floatarg dy,
 #endif
 {
 #if !PLUGDATA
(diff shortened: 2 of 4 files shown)


ben-wes:
last force push fixes crashes on cleanup. this is now basically a copy of the mechanisms in https://github.com/iem-projects/pd-iemguts/blob/main/src/receivecanvas.c

charlesneimog:
Hi @ben-wes, this is more of a question. 

Wouldn't it be more useful to implement functions that calculate whether the mouse is inside or outside the canvas using Lua? I say this because I understand that if we had a function in which the mouse position is tracked (visible from Lua; the proxy method would pass the mouse positions to a Lua function and there check if the mouse is inside or outside), it would be possible to implement mouse resize methods from Lua or overwrite this proxy function, which would allow extending the use of mouse positions in general.

Some resizing mouse methods would be good (maybe it already is possible, but I don't know about it) hehe.

PR #74, opened 2025-03-05
Implement svg rendering for pd-vanilla and plugdata

timothyschoen:
WIP implementation of svg rendering for pd-vanilla. Works decently well, but I need to clean up some things.

Implementation for purr-data and plugdata is also still needed, but that should be much easier.


timothyschoen:
Also added support for plugdata, combined with some plugdata-side changes: https://github.com/plugdata-team/plugdata/tree/pdlua-svg

Works quite well!

Screenshot 2025-03-06 at 02 33 20

timothyschoen:
Screenshot 2025-03-06 at 13 28 01

More advanced demo: the SVG tiger in plugdata and pure-data. In terms of rendering capabilities, this is done now!

One caveat to be noted is that spng depends on zlib, so we need to link against that. I'm not very happy with that, so I'll look for a way to prevent it. Perhaps stbi_image can do it?

Something that's good to know, is that you can do escaped multi-line strings in lua like this:

```
local svg = [[
 ... svg here ...
]]

```

What's also nice, is that you can use svg rendering to circumvent any limitations of the rendering engine. If you want to have gradients or path intersections, you can just generate or write the svg for it. It won't be as performant on pd-vanilla, but at least you can pretty much draw everything this way.

timothyschoen:
Now:
- Got rid of all sys_vgui in favour of pdgui_vmess
- Make macros for plugdata/purr-data consistent
- Got rid of zlib dependency by using stbi
- Added a demo to the helpfile:

Screenshot 2025-03-06 at 16 03 00


I think that covers everything :)

timothyschoen:
Force pushed so we don't have the big spng header in the commit history

timothyschoen:
Whoops, I accidentally pushed some stuff here that should be separate PRs. I'll send a new one...

PR #76, opened 2025-03-11
Fix for missing `glist_findrtext` in new Pd versions

ben-wes:
this uses the mechanism we already had in place for the multichannel handling.

uses the new `glist_getrtext` if present and otherwise falls back to the previous `glist_findrtext`

tested with Pd vanilla for official `0.55-2` build and current `master` including the change.

### DIFF
--- a/pdlua.c
+++ b/pdlua.c
@@ -59,8 +59,13 @@
 
 #include "pdlua_gfx.h"
 
+// function pointer type for signal_setmultiout (added with Pd 0.54)
 typedef void (*t_signal_setmultiout)(t_signal **, int); 
 static t_signal_setmultiout g_signal_setmultiout;
+
+// function pointer type for glist_getrtext/glist_findrtext (changes with Pd 0.56)
+typedef t_rtext *(*t_glist_rtext)(t_glist *, t_text *);
+static t_glist_rtext g_glist_getrtext;
 
 // This used to be in s_stuff.h, but not anymore since 0.55.1test1.
 int sys_trytoopenone(const char *dir, const char *name, const char* ext,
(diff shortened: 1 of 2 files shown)


ben-wes:
ah, sorry - had tim's changes in my branch here already! will reset to master and force push ...

EDIT: done

pdlua.c
```diff
@@ -59,8 +59,13 @@
 
 #include "pdlua_gfx.h"
 
-typedef void (*t_signal_setmultiout)(t_signal **, int); 
-static t_signal_setmultiout g_signal_setmultiout;
+// function pointer type for signal_setmultiout (added with Pd 0.54)
+typedef void (*t_signal_setmultiout_fn)(t_signal **, int); 
+static t_signal_setmultiout_fn g_signal_setmultiout;
```
umlaeute: i find it a bit annoying that this commit mixes the renaming of the typedef of an unrelated function with the fix for the missing `glist_findrtext`.

ben-wes: true - this is not related. i was reconsidering those type names with the additional function. but it should be fine without this change (struggling a bit with what's "good practice" there). updated the PR.

dvzrv:
@agraef this would be great to have in a new release! :pray:

ben-wes:
@dvzrv this is actually part of the current release on deken. i have been inactive here for quite a bit now... but we should certainly reflect these changes here, too. Will try to show some more initiative in the next weeks! thanks for the reminder. :)

dvzrv:
Thanks for the additional info!
I needed to backport this to be able to build against recent versions of pd on Arch Linux.

PR #78, opened 2025-06-20
replace sys_trytoopenone() with public open_via_path()

ben-wes:
not sure what the side effects here might be - but since we get warnings in Pd0.56 (probably coming in a few weeks) for every call of `sys_trytoopenone()`, it might be reasonable to use an approach with `open_via_path()`, which is part of Pd's public API ...?


ben-wes:
related to https://github.com/pure-data/pure-data/issues/2416 - and not sure this is the proper fix. but seems to work well for me.

PR #87, opened 2026-03-16
Combined PRs into large update

timothyschoen:
Work in progress to take some work away from @agraef for the next update.

I've merged most of the opened PRs into a new branch, resolving conflicts and adding missing plugdata implementations if needed. It includes:

https://github.com/agraef/pd-lua/pull/84: SVG rendering by @timothyschoen 
https://github.com/agraef/pd-lua/pull/82: Makefile improvement by @umlaeute 
https://github.com/agraef/pd-lua/pull/80: Clock deprecation fix by @ben-wes 
https://github.com/agraef/pd-lua/pull/79: Github actions runner fix by @ben-wes 
https://github.com/agraef/pd-lua/pull/77: Property menu support by @charlesneimog
https://github.com/agraef/pd-lua/pull/76: Fix compilation against new Pd version by @ben-wes 
https://github.com/agraef/pd-lua/pull/70: Add mouse enter/exit callbacks by @ben-wes 
https://github.com/agraef/pd-lua/pull/69: Fix path:close() documentation error by @ben-wes 
https://github.com/agraef/pd-lua/pull/66: Make some gfx arguments optional by @ben-wes 
https://github.com/agraef/pd-lua/pull/65: Add alignment feature to text drawing by @ben-wes
https://github.com/timothyschoen/pd-lua/pull/7: Fix missing return value by @KottV

On top of that:
- plugdata fixes:
  - Multi-instance now runs with a true independent lua instance per Pd instead, instead of using lua threads
  - Ported new property menu feature
  - Ported new text alignment feature
  - Ported mouse enter/exit callbacks
- Cleaned up properties implementation
    - Remove unused variables and unnecessary checks
    - Moved to separate file, since it works independently from pdlua graphics
    - Also allow properties on non-gui objects
    - Fix properties dialog title
    - Added float/int/combo property options
    - Fix properties button greying out when no properties() function is available
    - Make properties "apply" behaviour like other Pd property dialogs
- Added option to render images (using stb_image on pd)
- Added runtime choice between Lua and LuaJIT (chosen based on .pd_lua or .pd_luajit filename)
- Fixed Github Actions building with UCRT instead of MSVCRT (causing crashes on Windows when reading files)
- Fixed glitches when pdlua graphics objects get recreated on the same address as a previous pdlua object
- Fixed bug in rounded rectangle drawing on pd-vanilla
- Small code style fixes
- Updated to Lua 5.5
- Added documentation for new features

Planned:
- [ ] Port features to purr-data
   - [ ] SVG rendering
   - [ ] Image rendering
   - [ ] Mouse enter/exit
   - [ ] Text alignment
   - [x] Properties panels
       - Basics are implemented, though the UI could be better

I'm aware that this is gonna be a big set of changes. To ensure stability, I was hoping to get some assistance with testing from @ben-wes and @charlesneimog, once I get this out of the draft stage.


charlesneimog:
Hey @timothyschoen, amazing!!!! 

* Maybe would be great to update `lua` to version `5.5.0` too! 
* I also believe that the property menu support is not good. But I can look on it, to fix.

ben-wes:
thank you so much for taking the initiative here, @timothyschoen! i've been procrastinating on the next steps from my side quite a bit now, but i'll gladly contribute with extensive testing on macOS (and possibly Windows).

i was surprised that the builds here were still on pd0.55 (since the last deken upload from my side was probably 0.56 already and there should have been a PR for that as well) ... but looks like that's (almost) fixed now. the windows 64 build has some minor naming difference compared to other versions (Pd-0.56.2 instead of Pd-0.56-2 afaict from the logs). my fix for that in the workflow might be a bit ugly ...

timothyschoen:
> Hey @timothyschoen, amazing!!!!
> 
>     * Maybe would be great to update `lua` to version `5.5.0` too!
> 
>     * I also believe that the property menu support is not good. But I can look on it, to fix.

I believe Lua updates sometimes break backward compatibility, so I'm not sure if we can do that. Kind of depends on how much it breaks backward compatibility, I'll look into that.

I must admit, I didn't test the properties panel in pure-data yet, I went straight into porting it to plugdata, which does work:

Screenshot 2026-03-16 at 14 17 54

What's still broken about the properties panel? If you have time to help out, it is of course appreciated, I still have quite a bit of stuff to figure out :)

ben-wes:
> I believe Lua updates sometimes break backward compatibility

i admittedly don't know which changes in the latest Lua version we might profit from. but shouldn't the general idea be: pd-lua uses the latest lua version and if people need support for previous lua versions, they'd just use an older pd-lua version? although i see that this is not so simple in the plugdata context with pd-lua integrated... hmm ...

charlesneimog:
The main changes lua `5.5.0`.

``` 
Here are the main changes introduced in Lua 5.5. The [reference manual](https://www.lua.org/manual/5.5/contents.html) lists the [incompatibilities](https://www.lua.org/manual/5.5/manual.html#8) that had to be introduced.
Main changes

    declarations for global variables
    named vararg tables
    for-loop variables are read only
    floats are printed in decimal with enough digits to be read back correctly.
    more levels for constructors
    table.create
    utf8.offset returns also final position of character
    external strings (that use memory not managed by Lua)
    new functions luaL_openselectedlibs and luaL_makeseed
    major garbage collections done incrementally
    more compact arrays (large arrays use about 60% less memory)
    lua.c loads 'readline' dynamically
    static (fixed) binaries (when loading a binary chunk in memory, Lua can reuse its original memory in some of the internal structures)
    dump and undump reuse all strings
    auxiliary buffer reuses buffer when it creates final string
    
```
    
I was using lua 5.5 on my fork, and I didn't have problems.

timothyschoen:
Screenshot 2026-03-16 at 17 37 56

Screenshot 2026-03-16 at 17 38 32

Added number and combo properties now too!

And the API for properties has changed slightly:

```
function hello:properties(p)
    -- Checkboxes
    p:new_frame("First CheckBox", 2)
    p:add_check("Check Box 1", "updatecheckbox1", 1)
    p:add_check("Check Box 2", "updatecheckbox2", 1)

    -- Text inputs
    p:new_frame("First textinput", 2)
    p:add_text("Check text 1", "updatetext1", "test", 5)
    p:add_text("Check text 2", "updatetext2", "test", 5)

    -- Color picker
    p:new_frame("My Color Picker", 1)
    p:add_color("Background", "updatecolorbg")

    -- Number boxes
    -- add_number(label, callback, init, is_int, min, max)
    p:new_frame("Numbers", 2)
    p:add_number("Volume", "updatevolume", 0.5, 0, 0, 1)
    p:add_number("Voices", "updatevoices", 4, 1, 1, 16)

    -- Combobox
    -- add_combo(label, callback, init, options)
    p:new_frame("Mode Selection", 1)
    p:add_combo("Waveform", "updatewaveform", 1, { "Sine", "Square", "Saw", "Triangle" })
end
```

We no longer call functions on self to make sure that these functions are only available inside the properties() function. I've also made the function names closer to the rest of the pdlua API (a bit shorter and separating words with _)

charlesneimog:
I was remembering what I did, maybe it would be great if, instead of having a set button, I sent all the changes on Apply (instead of set) for text.

timothyschoen:
> I was remembering what I did, maybe it would be great if, instead of having a set button, I sent all the changes on Apply (instead of set) for text.

I agree, done with 50472be

Properties seem to be working quite well right now!

charlesneimog:
@timothyschoen maybe is missing a `pd.lua` commit, because you wrote this `function hello:properties(p)`, but on current commit (4158e718176deaa680a09ab9d67d53750434e249) we have this 

``` lua
pd._properties = function (object)
  local obj = pd._objects[object]
  if nil ~= obj and type(obj.properties) == "function" then
        obj:properties()
        return true
  else
        pd.post("not found")
        return false
  end
end
```

``` c
    // call _properties
    lua_getglobal(L, "pd");
    lua_getfield(L, -1, "_properties");

    lua_pushlightuserdata(L, pdlua);
    t_pdlua **ctx = lua_newuserdata(L, sizeof(t_pdlua*));
    *ctx = pdlua;

    luaL_getmetatable(L, "PropertiesContext");
    lua_setmetatable(L, -2);
```

I am curious eheheh

timothyschoen:
> @timothyschoen maybe is missing a `pd.lua` commit, because you wrote this `function hello:properties(p)`, but on current commit ([4158e71](https://github.com/agraef/pd-lua/commit/4158e718176deaa680a09ab9d67d53750434e249)) we have this
> 
> ```lua
> pd._properties = function (object)
>   local obj = pd._objects[object]
>   if nil ~= obj and type(obj.properties) == "function" then
>         obj:properties()
>         return true
>   else
>         pd.post("not found")
>         return false
>   end
> end
> ```
> 
> ```c
>     // call _properties
>     lua_getglobal(L, "pd");
>     lua_getfield(L, -1, "_properties");
> 
>     lua_pushlightuserdata(L, pdlua);
>     t_pdlua **ctx = lua_newuserdata(L, sizeof(t_pdlua*));
>     *ctx = pdlua;
> 
>     luaL_getmetatable(L, "PropertiesContext");
>     lua_setmetatable(L, -2);
> ```
> 
> I am curious eheheh

Whoops, missed that! Should be fixed now

charlesneimog:
Working like a charm now... I will do some test in complex objects like `bhack.voice`.

charlesneimog:
@timothyschoen I didn't understand this jit thing, this would be something just for `plugdata`, or pd uses would also be able to use this. I was searching about luajit, and mainly for audio with some gui would be great, I think for example for objects like `show~` of @ben-wes.  

I was doing some research with `luajit` on https://github.com/El-LOCUS-SOLUS/pd-lua, but this would be a compilation choice right?

timothyschoen:
> @timothyschoen I didn't understand this jit thing, this would be something just for `plugdata`, or pd uses would also be able to use this. I was searching about luajit, and mainly for audio with some gui would be great, I think for example for objects like `show~` of @ben-wes.
> 
> I was doing some research with `luajit` on https://github.com/El-LOCUS-SOLUS/pd-lua, but this would be a compilation choice right?

This user here has found a way to have both luajit and regular lua be available in the same build: https://x.com/Altius__/status/2032478814761271380 I believe it can work with Pd as well, it would be no different. Could indeed be great for dsp and gui performance.

I'm not exactly sure how they did it either, but they're planning to push it to github very soon, so we'll see if it can be merged in

timothyschoen:
- Image rendering support for many formats (using stb_image)
- Removed the "width" property for add_text because none of the other properties allow setting width
- Made colour property show visual feedback and preserve last state
- Added proper documentation for:
   - Image/svg rendering
   - Properties
   - Text alignment options
   - Default arguments

Screenshot 2026-03-17 at 11 50 09

Screenshot 2026-03-17 at 12 46 01


Note that the diff looks so large because of the added svg and image libraries, the actual changes to pdlua's code are not that big.

timothyschoen:
@charlesneimog It seems like right now, all properties apply immediately on change. We could either make it wait for the "apply" command, or we could replace the cancel/ok/apply buttons with just an ok button.

My preference might be to just have one button, I fear it would become quite complex otherwise

timothyschoen:
@agraef Regarding the purr-data implementation: SVGs, images, mouse enter/exit and text alignment should be trivial, just a matter of passing the data over to purr-data and handling it in JS.

There is already a basic implementation for purr-data external properties [here](https://github.com/agraef/purr-data/blob/7dc4f29c92952039d14d558155379432a6eea0fc/pd/nw/dialog_external.html), but it's missing a few features like:
- Colour picker
- Range and type limits for numbers
- "Frame" layouting, though this could be simplified into just having labels for each section like we do for plugdata

All in all, I think it's quite doable to port this over

charlesneimog:
> @charlesneimog It seems like right now, all properties apply immediately on change. We could either make it wait for the "apply" command, or we could replace the cancel/ok/apply buttons with just an ok button.
> 
> My preference might be to just have one button, I fear it would become quite complex otherwise
Here’s a polished version of your text:

Yes, I agree. I think saving everything in variables and sending it all at once is better. I also don’t like Pd’s `Apply` → `Ok` workflow, but I understand why it’s used.

If the user just closes the window, nothing will be applied, right? Do you think this might feel odd for users use to Pd? When I implemented it the first time, I found it could be a bit unexpected.

timothyschoen:
> > @charlesneimog It seems like right now, all properties apply immediately on change. We could either make it wait for the "apply" command, or we could replace the cancel/ok/apply buttons with just an ok button.
> > My preference might be to just have one button, I fear it would become quite complex otherwise
> > Here’s a polished version of your text:
> 
> Yes, I agree. I think saving everything in variables and sending it all at once is better. I also don’t like Pd’s `Apply` → `Ok` workflow, but I understand why it’s used.
> 
> If the user just closes the window, nothing will be applied, right? Do you think this might feel odd for users use to Pd? When I implemented it the first time, I found it could be a bit unexpected.

~Ah, so you would prefer it if the properties don't apply immediately? I'll work on that~

Done, properties now get sent only after clicking "Apply" or "OK", and are discarded with "Cancel".

umlaeute:
> ~~Ah, so you would prefer it if the properties don't apply immediately? I'll work on that~~
>
> Done, properties now get sent only after clicking "Apply\" or \"OK\", and are discarded with \"Cancel\".

I think this is a matter of design guidelines for different OSs (rather than personal preferences).

eg on macOSX, properties should apply *immediately* (whenever a value changes).
on Windows and Linux (even though there are no design guidelines imposed) the general expectation is to be able to tell the system, when you want to *apply* (by clicking a button).

that's the reason why Pd's dialogs look they do

timothyschoen:
> > ~Ah, so you would prefer it if the properties don't apply immediately? I'll work on that~
> > Done, properties now get sent only after clicking "Apply" or "OK", and are discarded with "Cancel".
> 
> I think this is a matter of design guidelines for different OSs (rather than personal preferences).
> 
> eg on macOSX, properties should apply _immediately_ (whenever a value changes). on Windows and Linux (even though there are no design guidelines imposed) the general expectation is to be able to tell the system, when you want to _apply_ (by clicking a button).
> 
> that's the reason why Pd's dialogs look they do

Thanks, that's good to know. So from my experimenting so far, I've found:

Windows/Linux:
- Only apply properties after clicking "Apply"
- Hitting return key applies and closes window

macOS:
- Color properties apply immediately when set
- Other properties apply on hitting return key, hitting return twice closes the window as well
- There is no apply button

Does that sound right?

timothyschoen:
I managed to build a version that combines both lua and luajit into one external:

https://github.com/timothyschoen/pd-lua/tree/luajit

Here's how it works:

Luajit side:
- Use macros to rename all pdlua_* functions to pdluajit_*
- Use the [compat53](https://github.com/lunarmodules/lua-compat-5.3) library to make some of the modern lua features we use available to luajit, which is normally based on Lua 5.1. This includes both a C header to expand the C API, and a Lua library which we inject on startup
- Define LUA_FILE_EXTENSION as ".pd_luajit"
- Then, include "pdlua.c" and link against luajit

Regular lua side:
- Use macros to prefix all lua functions with "lua54_"
- Define LUA_FILE_EXTENSION as ".pd_lua"
- Then, include both onelua.c and pdlua.c
- Call pdluajit_setup() at the end of pdlua_setup()

After doing that, it works perfectly: both get compiled into a single external. You decide which Lua gets used based on whether the lua file ends with .pd_lua or .pd_luajit

Here's a benchmark comparing the two:

```
pdlua:
-----------------------------------
empty loop           0.013680 sec
arithmetic           0.060547 sec
function calls       0.076619 sec
table access         0.044508 sec
math.sin             0.152804 sec

pdluajit:
-----------------------------------
empty loop           0.002288 sec   ->  6x faster
arithmetic           0.010526 sec   ->  6x faster
function calls       0.002264 sec   ->  33x faster
table access         0.012697 sec   ->  3.5x faster
math.sin             0.035742 sec    ->  4.25x faster
```

Personally I'd be in favour of merging this, since the performance difference is enormous. This also makes pdlua much more interesting for DSP purposes. I understand that it's a big change though, and the build system is admittedly a bit hacky.

charlesneimog:
This is really really cool, I am trying to compile it on arch linux, I am getting this: 

```
++++ info: making luas/luajit.pd_linux.o in lib pdlua
cc -DPD -I "/usr/include/pd"  -DUNIX  -fPIC -DMAKE_LIB -Iluas/luajit/src -DLUA_USE_LINUX -Iluas/lua-compat-5.3 -DPDLUA_VERSION="plugdata/0.10.2-482-gce59fff" -Wall -Wextra -Wshadow -Winline -Wstrict-aliasing -O3 -ffast-math -funroll-loops -fomit-frame-pointer -march=core2 -mfpmath=sse -msse -msse2 -msse3 -o luas/luajit.pd_linux.o -c luas/luajit.c
luas/luajit.c: In function ‘preload_compat53’:
luas/luajit.c:19:37: error: ‘luas_lua_compat_5_3_compat53_init_lua’ undeclared (first use in this function); did you mean ‘luas_lua_compat_5_3_compat53_file_mt_lua’?
   19 |     luaL_loadbuffer(L, (const char*)luas_lua_compat_5_3_compat53_init_lua,
      |                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                     luas_lua_compat_5_3_compat53_file_mt_lua
luas/luajit.c:19:37: note: each undeclared identifier is reported only once for each function it appears in
luas/luajit.c:20:24: error: ‘luas_lua_compat_5_3_compat53_init_lua_len’ undeclared (first use in this function); did you mean ‘luas_lua_compat_5_3_compat53_file_mt_lua_len’?
   20 |                        luas_lua_compat_5_3_compat53_init_lua_len, "compat53");
      |                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                        luas_lua_compat_5_3_compat53_file_mt_lua_len
In file included from luas/../pdlua.c:117,
```

Some ideia?

EDIT: Now works!!

timothyschoen:
> This is really really cool, I am trying to compile it on arch linux, I am getting this:
> 
> ```
> ++++ info: making luas/luajit.pd_linux.o in lib pdlua
> cc -DPD -I "/usr/include/pd"  -DUNIX  -fPIC -DMAKE_LIB -Iluas/luajit/src -DLUA_USE_LINUX -Iluas/lua-compat-5.3 -DPDLUA_VERSION="plugdata/0.10.2-482-gce59fff" -Wall -Wextra -Wshadow -Winline -Wstrict-aliasing -O3 -ffast-math -funroll-loops -fomit-frame-pointer -march=core2 -mfpmath=sse -msse -msse2 -msse3 -o luas/luajit.pd_linux.o -c luas/luajit.c
> luas/luajit.c: In function ‘preload_compat53’:
> luas/luajit.c:19:37: error: ‘luas_lua_compat_5_3_compat53_init_lua’ undeclared (first use in this function); did you mean ‘luas_lua_compat_5_3_compat53_file_mt_lua’?
>    19 |     luaL_loadbuffer(L, (const char*)luas_lua_compat_5_3_compat53_init_lua,
>       |                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>       |                                     luas_lua_compat_5_3_compat53_file_mt_lua
> luas/luajit.c:19:37: note: each undeclared identifier is reported only once for each function it appears in
> luas/luajit.c:20:24: error: ‘luas_lua_compat_5_3_compat53_init_lua_len’ undeclared (first use in this function); did you mean ‘luas_lua_compat_5_3_compat53_file_mt_lua_len’?
>    20 |                        luas_lua_compat_5_3_compat53_init_lua_len, "compat53");
>       |                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>       |                        luas_lua_compat_5_3_compat53_file_mt_lua_len
> In file included from luas/../pdlua.c:117,
> ```
> 
> Some ideia?

Can you try a clean build now? I was still messing around a bit to get the github actions to work on every platform, but it should be good now :)

charlesneimog:
@timothyschoen, on properties yet! (pdluajit works here). I was trying to do this

``` lua
function lplot:properties(p)
	local i = 0
	p:new_frame("Colors", #self.draws)

	for k, v in pairs(self.draws) do
		local method = "update_color_" .. i
		self[method] = function(obj, color)
			obj.draws[k].color = color[1]
		end
		p:add_color(k, method, v.color)
		i = i + 1
	end
end
```

I believe you need to add a 

```c
    pdlua->properties.property_count++;
```

on `static int pdlua_properties_addcolor(lua_State *L)` for Pure Data.

without this we have

``` 
(Tcl) UNHANDLED ERROR: window name "color0" already exists in parent
    while executing
"{frame} {.0x7f10243cc200.main.frame1.content.color0} "
    ("uplevel" body line 41)
    invoked from within
"uplevel #0 $docmds"
```

timothyschoen:
> @timothyschoen, on properties yet! (pdluajit works here). I was trying to do this
> 
> ```lua
> function lplot:properties(p)
> 	local i = 0
> 	p:new_frame("Colors", #self.draws)
> 
> 	for k, v in pairs(self.draws) do
> 		local method = "update_color_" .. i
> 		self[method] = function(obj, color)
> 			obj.draws[k].color = color[1]
> 		end
> 		p:add_color(k, method, v.color)
> 		i = i + 1
> 	end
> end
> ```
> 
> I believe you need to add a
> 
> ```c
>     pdlua->properties.property_count++;
> ```
> 
> on `static int pdlua_properties_addcolor(lua_State *L)` for Pure Data.
> 
> without this we have
> 
> ```
> (Tcl) UNHANDLED ERROR: window name "color0" already exists in parent
>     while executing
> "{frame} {.0x7f10243cc200.main.frame1.content.color0} "
>     ("uplevel" body line 41)
>     invoked from within
> "uplevel #0 $docmds"
> ```

Good catch, fixed that for both branches.

timothyschoen:
Added documentation as well:

Screenshot 2026-03-18 at 19 50 33


Another cool thing LuaJIT can do, is interface with C functions:

```
local ffi = require("ffi")

ffi.cdef[[
void post(const char *fmt, ...);
float sys_getsr(void);
double clock_getlogicaltime(void);
]]

ffi.C.post("=== Pd via LuaJIT FFI ===")
ffi.C.post("Samplerate: " .. tostring(ffi.C.sys_getsr()))
ffi.C.post("Time: " ..  tostring(ffi.C.clock_getlogicaltime()))
```

timothyschoen:
Merged luajit into this branch, since it just works and provides valuable features

timothyschoen:
@ben-wes: 121fe92 fixes the glitch we encountered during the live demo at LAC:

Screenshot 2026-03-20 at 14 40 52

The issue was: when an object gets recreated at the same address, it can run repaint() before running gfx_initialize, leading it to draw the content of the object that previously lived there. In this case, the first object had 4 layers and the second had 3, se the last layer would get stuck on canvas

ben-wes:
> @ben-wes: [121fe92](https://github.com/agraef/pd-lua/commit/121fe922593226a7423a5897b4519dce8648547d) fixes the glitch we encountered during the live demo at LAC

incredible, @timothyschoen! ... i still have to test all these new features. thank you!

timothyschoen:
I've tested this quite thoroughly on macOS/Windows/Linux with pure-data and plugdata, everything seems to work now. I'd say it's ready for further testing (if @ben-wes, @charlesneimog and @agraef have time soon). 

Apart from all the new features, there's a few bugs fixed here too, it should be more stable than the last release.

ben-wes:
... quickly tested my everyday object `show~` as luajit on 4f893ac. i'm really impressed that this just works (and also quite a bit faster - even with vanilla's drawing restrictions)!

the dualsense controller display also works. tomorrow, i'll have a more detailed look at the help files. thank you so much for your work on this!

timothyschoen:
Ok, actually done for real now :)

Screenshot 2026-03-21 at 21 32 33

Here's a test suite you can use to verify all the new features work, both under Lua 5.5 and LuaJIT:

[test-pdlua.zip](https://github.com/user-attachments/files/26161621/test-pdlua.zip)

ben-wes:
@timothyschoen one minor detail i realized when using the luajit version lately: obviously, messages like `reload` to `pdluax` don't work in this case. i assume that's expected - but not sure. i got used to that for updating lua objects during development. so for luajit, this mechanism is not intended?

